PR #7, opened 2024-02-11
feature/webrtc stats

amitovadia-did:
Gather webrtc stats and send report to callback

report includes segmented stats about each second of the video, and takes the 5 highest offenders in regards to packet loss.

### DIFF
--- a/demo-app/app.tsx
+++ b/demo-app/app.tsx
@@ -59,7 +59,7 @@ export function App() {
         }
     }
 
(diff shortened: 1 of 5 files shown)


benyael15 (approved):
are we sure of the triggers of events?
StreamEvents.StreamStarted - is video(talk) started
StreamEvents.StreamDone - is when session is dead, no? or perhaps here "done" mean idle?
lets make sure when in office that its what we wanted
other then that - great

PR #25, opened 2024-03-27
Add Connect + change rate

HaMMeRSI:
…reateSocketManager


src/createAgentManager.ts
```diff
@@ -106,136 +132,161 @@ export function getAgent(agentId: string, auth: Auth, baseURL?: string): Promise
...
+
+            const response = await agentsApi.chat(agentInstance.id, items.chat.id, {
+                sessionId: items.streamingManager.sessionId,
+                streamId: items.streamingManager.streamId,
+                messages: items.messages,
+                append_chat,
+            });
+
+            items.messages.push({
+                id: getRandom(),
```
benyael15, line 228: if user dont send it throw the API?
is the API create ID for message in BE - if not please add

benyael15 (approved):

PR #32, opened 2024-04-07
ask for stream warmup

osimhi213:
verified that stream_warmup is sent on streams post request

### DIFF
--- a/src/types/stream/api/clip.ts
+++ b/src/types/stream/api/clip.ts
@@ -48,6 +48,13 @@ export interface CreateClipStreamRequest {
      * @default auto
      */
     compatibility_mode?: CompatibilityMode;
+    /**
+     * Whether to stream wamrup video on the connection.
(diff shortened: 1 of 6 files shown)


src/types/stream/api/clip.ts
```diff
@@ -48,6 +48,13 @@ export interface CreateClipStreamRequest {
      * @default auto
      */
     compatibility_mode?: CompatibilityMode;
+    /**
+     * Whether to stream wamrup video on the connection.
+     * If set to true, will stream a warmup video when connection is established.
+     * At the end of the warmup video, a message containing "stream/ready" will be sent on the data channel.
+     * @default false
```
amitovadia-did, line 55: remove default false if we use default true

amitovadia-did (approved):

src/types/stream/api/clip.ts
osimhi213: was removed. thx

PR #33, opened 2024-04-09
Zoom-out for SP - phase 2 using stitch

o-bieliaiev:


src/createAgentManager.ts
```diff
@@ -47,10 +47,17 @@ function getAgentStreamArgs(agent: Agent): CreateStreamOptions {
...
+    if(agent.presenter.type === VideoType.Talk && agent.presenter.stitch) {
+        config = {
+            stitch: agent.presenter.stitch
+        }
+    }
     return {
         videoType: VideoType.Talk,
         source_url: agent.presenter.source_url,
         stream_warmup: true,
+        config: config && Object.keys(config).length > 0 ? config : undefined
```
o-bieliaiev, line 60: Make this to not send config object when stitch is false, to avoid any possible issue inside create stream endpoint

PR #34, opened 2024-04-10
Feature/agent manager chat mode

almog100:
Chat with text only  mode: chatWithoutStream


src/types/entities/agents/manager.ts
```diff
@@ -102,12 +102,16 @@ export interface AgentManager {
      * Method to close all connections with agent, stream and web socket
      */
     disconnect: () => Promise<void>;
+    /**
```
benyael15, line 105: wouldnt it be easier and cleaner to do some like

chat: (userMessage: string, chatMode: chat-mpode,append_chat?: boolean)

within the same methos?

src/createAgentManager.ts
```diff
@@ -252,7 +255,70 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
 
             analytics.track('agent-chat', { event: 'disconnect', chatId: items.chat?.id, agentId: agentInstance.id });
         },
-        async chat(userMessage: string, append_chat: boolean = false, enforceTextOnly: boolean = false) {
+        async chatWithoutStream(userMessage: string) {
+            try {
```
benyael15, line 259: add chat_mode to all MP analytics

src/types/entities/agents/manager.ts
almog100: we wanted to handle chatMode only in agents-ui, so we separate everything in agent-sdk so things won't get mixed up and only in agents-ui I have this differentiation by chatMode.

benyael15: gotchta - this is sdk :)

infra/build.js
```diff
@@ -36,6 +36,8 @@ try {
     console.log('start copy');
     const root = path.resolve(import.meta.url, '../../').split(':')[1];
     const dist = path.resolve(root, './dist');
+    const packageJson = path.resolve(root, './package.json');
+    console.log(packageJson);
```
benyael15, line 40: remove log

infra/build.js
```diff
@@ -52,6 +54,7 @@ try {
 
     console.log('Copying new package');
     fs.cpSync(dist, embeddedSdk, { recursive: true });
+    fs.copyFileSync(packageJson, path.resolve(embeddedSdk, '../package.json'));
 } catch (e) {
-    console.log('Copy failed');
+    console.log('Copy failed', e);
```
benyael15, line 59: remove

benyael15 (approved):

infra/build.js
HaMMeRSI: this is the build script

PR #40, opened 2024-04-29
Consume video events

nhantd:


src/createAgentManager.ts
```diff
@@ -184,6 +185,9 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
                 }
 
                 options.callbacks.onNewMessage?.(items.messages);
+            } else if ('videoStartTime' in data) {
```
benyael15, line 188: why not condition bu the event name?

src/createAgentManager.ts
```diff
@@ -163,27 +163,34 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
...
+                if (event === ChatProgress.Complete) {
                     analytics.track('agent-message-received', { messages: items.messages.length });
                 }
 
                 options.callbacks.onNewMessage?.(items.messages);
+            } else if ([StreamEvents.StreamVideoCreated, 
+                        StreamEvents.StreamVideoDone, 
+                        StreamEvents.StreamVideoError, 
+                        StreamEvents.StreamVideoRejected].includes(event as StreamEvents)) {
+                // Stream video event
```
benyael15, line 192: i think we need to add event: event to the MP payload
where is the stop/start?

benyael15: like this
analytics.track('agent-video', { event: event, ...message});

benyael15 (changes requested):

benyael15 (approved):
just make sure the events are in the payload
Stop/Start

PR #47, opened 2024-05-07
adding stream resolution option in create stream

osimhi213:
- adding stream resolution configuration for sdk users
- agents use 1080 for stitch as default

### DIFF
--- a/src/types/entities/agents/manager.ts
+++ b/src/types/entities/agents/manager.ts
@@ -85,6 +85,11 @@ export interface AgentManagerOptions {
      */
     distinctId?: string;
     mode?: ChatMode;
+    /**
+     * Desired stream resolution for the session
+     * @minimum 0
+     * @maximum 1080
+     */
     streamResolution?: number;
 }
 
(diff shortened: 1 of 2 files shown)


src/types/entities/agents/manager.ts
```diff
@@ -85,6 +85,12 @@ export interface AgentManagerOptions {
      */
     distinctId?: string;
     mode?: ChatMode;
+    /**
+     * Desired stream resolution for the session
+     * @minimum 0
+     * @maximum 1080
+     */
+    streamResolution?: number;
```
HaMMeRSI, line 93: Can we support all resolutions?
do we want to?
maybe limit to
144 | 360 | 720 | 1080
like youtube?

HaMMeRSI (approved):

src/types/entities/agents/manager.ts
osimhi213: there is a limit of minimum resolution but we don't acutally limit it.
maybe let's keep it as number between 0 and 1080 cause the api supports all and then think of limit it to what you were sasying.
is that ok?  @HaMMeRSI

PR #49, opened 2024-05-12
relying on stream/started event to start streaming

osimhi213:
- rather than webrtc stats
- to avoid first sentence cut-off
- in agents-ui we set opacity to 1 when it actually start playing so no risk 
https://app.asana.com/0/1207192568376365/1207237906837102/f


HaMMeRSI (approved):
Dont we want to remove the logic entirly?

osimhi213:
No I don't think so

PR #53, opened 2024-05-19
setting stats interval to 100ms

osimhi213:
- sensitive to stream start
- fixing the cutoff bug
- https://app.asana.com/0/1207192568376365/1207237906837102/f
- @benyael15 don't merge yet, I want to check with you the stats to mixpanel..

### DIFF
--- a/src/createStreamingManager.ts
+++ b/src/createStreamingManager.ts
@@ -77,10 +77,10 @@ function pollStats(peerConnection: RTCPeerConnection, onVideoStateChange, analyt
                 isStreaming = true;
             }
 
-        } else {
+        } else if (isStreaming) {
             notReceivingNumIntervals++;
 
-            if (isStreaming && notReceivingNumIntervals >= notReceivingIntervalsThreshold) {
+            if (notReceivingNumIntervals >= notReceivingIntervalsThreshold) {
                 onVideoStateChange?.(StreamingState.Stop);
 
                 isStreaming = false;
--- a/src/createStreamingManager.ts
+++ b/src/createStreamingManager.ts
@@ -41,7 +41,7 @@ function createVideoStatsAnalyzer() {
     let lastBytesReceived = 0;
 
     return (stats: RTCStatsReport) => {
-        for (const report of stats.values()) {
+        stats.forEach(report => {
             if (report.type === 'inbound-rtp' && report.kind === 'video') {
                 if (report) {
                     const currBytesReceived = report.bytesReceived;
@@ -52,7 +52,7 @@ function createVideoStatsAnalyzer() {
                     return isReceiving
                 }
             }
-        };
+        });
 
         return false;
     };
@@ -64,7 +64,7 @@ function pollStats(peerConnection: RTCPeerConnection, onVideoStateChange, analyt
 
     let notReceivingNumIntervals = 0;
     let isStreaming = false;
-    
+
     const isReceivingBytes = createVideoStatsAnalyzer();
 
     return setInterval(async () => {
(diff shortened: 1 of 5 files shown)


src/createStreamingManager.ts
```diff
@@ -37,57 +37,58 @@ function mapConnectionState(state: RTCIceConnectionState): ConnectionState {
...
-                        if (!isPlaying && isPlayingFalseNumIntervals === 2) {
-                            onVideoStateChange?.(StreamingState.Stop, videoStatsReport);
-                        }
-                    }
-                    videoStats.push(report);
+function createVideoStatsAnalyzer() {
+    let lastBytesReceived = 0;
+
+    return (stats: RTCStatsReport) => {
+        stats.forEach(report => {
```
HaMMeRSI, line 44: can we change it to 
for (const report of stats) ?

src/createStreamingManager.ts
```diff
@@ -37,57 +37,58 @@ function mapConnectionState(state: RTCIceConnectionState): ConnectionState {
...
+            if (report.type === 'inbound-rtp' && report.kind === 'video') {
+                if (report) {
```
HaMMeRSI, line 46: if (report.type === 'inbound-rtp' && report.kind === 'video') {
                if (report) {

the order doesn't make sense

src/createStreamingManager.ts
```diff
@@ -37,57 +37,58 @@ function mapConnectionState(state: RTCIceConnectionState): ConnectionState {
...
+                    const currBytesReceived = report.bytesReceived;
+                    const isReceiving = currBytesReceived - lastBytesReceived > 0;
+
+                    lastBytesReceived = currBytesReceived
+
+                    return isReceiving
```
HaMMeRSI, line 52: ;

src/createStreamingManager.ts
```diff
@@ -37,57 +37,58 @@ function mapConnectionState(state: RTCIceConnectionState): ConnectionState {
...
+    let lastBytesReceived = 0;
+
+    return (stats: RTCStatsReport) => {
+        stats.forEach(report => {
+            if (report.type === 'inbound-rtp' && report.kind === 'video') {
+                if (report) {
+                    const currBytesReceived = report.bytesReceived;
+                    const isReceiving = currBytesReceived - lastBytesReceived > 0;
+
+                    lastBytesReceived = currBytesReceived
```
HaMMeRSI, line 50: ;

src/createStreamingManager.ts
```diff
@@ -37,57 +37,58 @@ function mapConnectionState(state: RTCIceConnectionState): ConnectionState {
...
+            notReceivingNumIntervals = 0;
+
+            if (!isStreaming) {
+                onVideoStateChange?.(StreamingState.Start);
+
+                isStreaming = true;
+            }
+
+        } else {
+            notReceivingNumIntervals++;
```
HaMMeRSI, line 83: you always increase it no?

HaMMeRSI (changes requested):

src/createStreamingManager.ts
osimhi213: I kept the original order, you're right. I'll add it to the condition

src/createStreamingManager.ts
osimhi213: as long as not receiving no ? 
I zero it overhere:
if (isReceivingBytes(stats)) {
            notReceivingNumIntervals = 0;

            if (!isStreaming) {

                isStreaming = true;
            }

        }

osimhi213: done

HaMMeRSI (approved):

PR #55, opened 2024-05-22
Reconnect with new stream

nhantd:
https://app.asana.com/0/1207192568376365/1207237906837080/f

When stream is invalid, then .chat() function should retry to create new stream id, and send the chat again. This is to avoid user/agent falling into message not reponsed.

https://github.com/de-id/agents-sdk/assets/6271974/8ec56289-e463-4a8a-bb06-9308a6b6ce8d

### DIFF
--- a/src/createAgentManager.ts
+++ b/src/createAgentManager.ts
@@ -271,6 +271,22 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
         }
     }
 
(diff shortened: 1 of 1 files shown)


benyael15 (commented):
how does it garrantee that the message sent after recoonnection wont appear twice as its today?

benyael15 (approved):

PR #61, opened 2024-06-04
Add latency to mixpanel

voldemortman:


benyael15 (changes requested):
this should go to agent-video: start event as `upstreamLatency`

voldemortman:
Changes aren't needed in agents-sdk in order to get upstream latency and template

PR #79, opened 2024-07-17
Add status to agent type

voldemortman:


src/types/entities/agents/agent.ts
```diff
@@ -42,6 +51,7 @@ export interface Agent {
     logo?: string;
     preview_url?: string;
     owner_id?: string;
+    status?: AgentStatus;
```
dor-eitan, line 54: I think agent always have a status

dor-eitan (approved):

PR #80, opened 2024-07-21
dont send the matches back to chat

HaMMeRSI:
sends tons of unnecessary data
https://d-id-api.sentry.io/issues/4925552627/events/27263caee342451db7ea1933a75a0c3e/


benyael15 (approved):
so the issue is that the sdk sent this aggregation to chat-api within any new message?

PR #81, opened 2024-07-22
Fix ConnectionState updates

nivzelber:


src/createAgentManager.ts
```diff
@@ -291,6 +291,8 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
     };
 
     async function connect(newChat: boolean) {
+        options.callbacks.onConnectionStateChange?.(ConnectionState.Connecting);
```
nivzelber, line 294: @HaMMeRSI is it safe to set this on our own, even when ice does it? can the double assignment cause issues?

osimhi213 (approved):

osimhi213 (approved):

PR #86, opened 2024-08-22
Pass persist to create chat

voldemortman:
- [Asana](https://app.asana.com/0/1207922108504705/1207861165820789/f)

### DIFF
--- a/src/api/agents.ts
+++ b/src/api/agents.ts
@@ -25,8 +25,8 @@ export function createAgentsApi(
         update(id: string, payload: AgentPayload, options?: RequestInit) {
             return client.patch<Agent>(`/${id}`, payload, options);
         },
-        newChat(agentId: string, options?: RequestInit) {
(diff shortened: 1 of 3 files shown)


src/api/agents.ts
```diff
@@ -25,8 +25,8 @@ export function createAgentsApi(
         update(id: string, payload: AgentPayload, options?: RequestInit) {
             return client.patch<Agent>(`/${id}`, payload, options);
         },
-        newChat(agentId: string, options?: RequestInit) {
-            return client.post<Chat>(`/${agentId}/chat`, undefined, options);
+        newChat(agentId: string, payload: { persist: boolean }, options?: RequestInit) {
```
HaMMeRSI, line 28: call it options

HaMMeRSI (approved):

src/api/agents.ts
voldemortman: There already is options in the function signature and it is the actual payload of the request

PR #98, opened 2024-11-06
Feature/stream greeting

voldemortman:


src/api/clipStream.ts
```diff
@@ -25,6 +25,7 @@ export function createApi(
                 compatibility_mode: options.compatibility_mode,
                 stream_warmup: options.stream_warmup,
                 session_timeout: options.session_timeout,
+                greeting: options.greeting,
```
benyael15, line 28: call it "stream_greeting"

src/createAgentManager.ts
```diff
@@ -42,13 +42,14 @@ interface ChatEventQueue {
     answer?: string;
 }
 
-function getAgentStreamArgs(agent: Agent, options?: AgentManagerOptions): CreateStreamOptions {
```
benyael15, line 45: greeting should be then part of "stream options" objects and type
better scalable of SDK api is to enlarge an exisintg object with more fields then adding method params

benyael15 (changes requested):
better extendability technique

src/createAgentManager.ts
voldemortman: It is already part of the agent manager options (Which is client facing). The greeting param is the text to be streamed. I had to include the text to be streamed in order to deal with agents that don't have greetings set in the db. All of the default greetings are not stored in the db.

benyael15: i meant i want it to be part of options
instead of
(agent, chat, options, greetings)
i want
(agent, chat, options) where options is :{...prevOptions, greetings}

voldemortman: It is part of options. In the options the user can pass whether or not they want the greeting to be streamed or not. The greeting field here is the actual greeting to be streamed. This function isn't exposed to the users. Do you want the user to be able to control the actual message that is being sent?

benyael15: maybe its redundent then? the actual text is in agent (when creating the agent) - its enough IMO
otherwise fallback to "my name is...."

voldemortman: I just didn't want the default greeting message to be in serverless as well. I preferred it all to be agents-sdk, and have the sdk pass the message to serverless. I can change it if you don't think it's a problem

benyael15 (approved):

PR #101, opened 2024-12-17
Feature/sync greeting and ready

voldemortman:


src/createStreamingManager.ts
```diff
@@ -76,7 +85,11 @@ function pollStats(peerConnection: RTCPeerConnection, onVideoStateChange) {
 
             if (!isStreaming) {
                 onVideoStateChange?.(StreamingState.Start);
-
+                if (streamsCount >= streamsBeforeReady && !getIsConnected()) {
+                    console.log('connecting from video stream')
+                    connect();
```
benyael15, line 90: why connect?

benyael15 (commented):
review with @HaMMeRSI

src/createStreamingManager.ts
benyael15: call is connectionCallback
its more clear when connect that at first i tought is the actul connetion api - but its not
it is just a callback and flag

src/createAgentManager.ts
```diff
@@ -118,6 +118,7 @@ function initializeStreamAndChat(
                 callbacks: {
                     ...options.callbacks,
                     onConnectionStateChange: async state => {
+                        options.callbacks.onConnectionStateChange?.(state);
```
HaMMeRSI, line 121: whyd you moved it here?

connected is only happens after new chat started

src/createAgentManager.ts
```diff
@@ -326,7 +324,7 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
         async function connectWithRetry() {
             for (let attempt = 1; attempt <= maxRetries; attempt++) {
                 try {
-                    return await initializeStreamAndChat(agentInstance, options, agentsApi, analytics, items.chat, greeting);
+                    return await initializeStreamAndChat(agentInstance, options, agentsApi, analytics, items.chat, ...(newChat ? [greeting] : []));
```
HaMMeRSI, line 327: aint there a prettier way of doing this?
move this logic to getGreeting()?

src/createStreamingManager.ts
```diff
@@ -58,13 +59,21 @@ function createVideoStatsAnalyzer() {
     };
 }
 
-function pollStats(peerConnection: RTCPeerConnection, onVideoStateChange) {
+function pollStats(
+    peerConnection: RTCPeerConnection,
+    onVideoStateChange,
+    warmup: boolean = false,
+    getIsConnected: () => boolean,
+    connect: () => void) {
```
HaMMeRSI, line 67: onConnect

HaMMeRSI (changes requested):

PR #104, opened 2025-01-05
Bugfix/mixpanel events

almog100:
- Description
  Enriched all mixpanel events in sdk with: language, presenter id (for clip presenters), agents ui version and window size.


src/services/mixpanel.ts
```diff
@@ -77,9 +77,13 @@ export function initializeAnalytics(config: AnalyticsOptions): Analytics {
                                 time: Date.now(),
                                 $insert_id: this.getRandom(),
                                 origin: window.location.href,
-                                'Screen Height': window.screen.height || window.innerWidth,
-                                'Screen Width': window.screen.width || window.innerHeight,
+                                'Screen Height': window.screen.height,
+                                'Screen Width': window.screen.width,
```
almog100, line 81: fixed this wrong fallback to innerWidth.

src/types/global.d.ts
```diff
@@ -0,0 +1,8 @@
+declare global {
+    interface Window {
+        DID_AGENTS_API?: boolean;
+        DID_AGENTS_UI_VERSION?: string;
```
benyael15, line 4: i dont think this needs to live in SDK
man users are using it regardless of UI component - inject it from the UI without sdk knowing its his

src/services/mixpanel.ts
```diff
@@ -77,9 +77,13 @@ export function initializeAnalytics(config: AnalyticsOptions): Analytics {
...
+                                'Window Size': `${window.innerWidth}x${window.innerHeight}`,
                                 'User Agent': navigator.userAgent,
+                                ...(presenter.type === 'clip' && { 'Presenter Id': presenter.presenter_id }),
+                                Language: navigator.language,
```
benyael15, line 85: missing presenter type

benyael15, line 85: Language of the agent i think its the purpose
the same lang set in to the speech recognition

src/services/mixpanel.ts
```diff
@@ -77,9 +77,13 @@ export function initializeAnalytics(config: AnalyticsOptions): Analytics {
                                 time: Date.now(),
                                 $insert_id: this.getRandom(),
                                 origin: window.location.href,
-                                'Screen Height': window.screen.height || window.innerWidth,
-                                'Screen Width': window.screen.width || window.innerHeight,
+                                'Screen Height': window.screen.height,
+                                'Screen Width': window.screen.width,
+                                'Window Size': `${window.innerWidth}x${window.innerHeight}`,
```
benyael15, line 82: window_size - means the component size (as its resizeable inv2) - should be injected from UI

PR #106, opened 2025-01-15
feature/support-stt-token-generation

nivzelber:
Expose get stt token request, and current chat of the agent

### DIFF
--- a/src/createAgentManager.ts
+++ b/src/createAgentManager.ts
@@ -422,11 +422,7 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
     return {
         agent: agentInstance,
         starterMessages: agentInstance.knowledge?.starter_message || [],
-        getSTTToken: (options?: RequestInit) => {
-            return items.chat?.id
-                ? agentsApi.getSTTToken(agentInstance.id, items.chat.id, options)
-                : Promise.resolve(undefined);
-        },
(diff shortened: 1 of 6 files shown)


src/createAgentManager.ts
```diff
@@ -415,6 +422,11 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
     return {
         agent: agentInstance,
         starterMessages: agentInstance.knowledge?.starter_message || [],
+        getSTTToken: (options?: RequestInit) => {
```
HaMMeRSI, line 425:
```js
async (options?: RequestInit) => {
            if(items.chat?.id) {
                return agentsApi.getSTTToken(agentInstance.id, items.chat.id, options)
    }
}
```

HaMMeRSI: + dont provide the request options.
we don't anywhere else + there is nothing to change

HaMMeRSI (approved):

src/createAgentManager.ts
nivzelber: so like that?

```ts
async () => items.chat?.id ? agentsApi.getSTTToken(agentInstance.id, items.chat.id) : undefined
}

HaMMeRSI: no, leave trenary, you dont need to resolve undefined if function is async, its auto

PR #111, opened 2025-01-26
client side disconnect on any error

osimhi213:
- disconnect if timeout occurred and streaming manager has been initialized 
- increase timeout to 30sec
- disconnect on any webrtc connection establishment error after creating a stream

### DIFF
--- a/src/createAgentManager.ts
+++ b/src/createAgentManager.ts
@@ -27,7 +27,7 @@ import { Analytics, initializeAnalytics } from './services/mixpanel';
 import { getAnaliticsInfo, getStreamAnalyticsProps } from './utils/analytics';
 
 let messageSentTimestamp = 0;
-const connectionRetryTimeoutInMs = 20 * 1000; // 20 seconds
+const connectionRetryTimeoutInMs = 30 * 1000; // 30 seconds
 
 interface AgentManagerItems {
     chat?: Chat;
@@ -107,8 +107,11 @@ function initializeStreamAndChat(
     return new Promise<{ chat?: Chat; streamingManager: StreamingManager<CreateStreamOptions> }>(
         async (outerResolve, outerReject) => {
             messageSentTimestamp = 0;
+            let streamingManager: StreamingManager<CreateStreamOptions> | null = null;
 
-            const timeoutId = setTimeout(() => {
+            const timeoutId = setTimeout(async () => {
+                await streamingManager?.disconnect().catch(e => console.error('Error during disconnect:', e));
+
                 reject(new Error('Could not connect'));
             }, connectionRetryTimeoutInMs);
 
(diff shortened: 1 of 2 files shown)


src/createAgentManager.ts
```diff
@@ -107,8 +107,11 @@ function initializeStreamAndChat(
     return new Promise<{ chat?: Chat; streamingManager: StreamingManager<CreateStreamOptions> }>(
         async (outerResolve, outerReject) => {
             messageSentTimestamp = 0;
+            let streamingManager: StreamingManager<CreateStreamOptions> | null = null;
+
+            const timeoutId = setTimeout(async () => {
```
nivzelber, line 112: move this `setTimeout` call below `createStreamingManager` call
1. You will start the timeout only after creating the manager (as opposed to now - where you wait 30 sec from the call to create manager, instead of 30 sec from the actual creation time)
1. You won't have to declare `let streamingManager...`

src/createStreamingManager.ts
```diff
@@ -139,63 +139,71 @@ export async function createStreamingManager<T extends CreateStreamOptions>(
...
-    await startConnection(streamIdFromServer, sessionClientAnswer, session_id);
-    log('start connection OK');
+        await startConnection(streamIdFromServer, sessionClientAnswer, session_id);
+        log('start connection OK');
+    }
+
+    try {
+        await establishWebRtcConnection();
+    } catch {
+        await close(streamIdFromServer, session_id).catch(_ => { });
```
nivzelber, line 205: Should we retry here? Or are we OK with the server holding the streaming until we reach the server-side timeout?

src/createStreamingManager.ts
```diff
@@ -233,9 +241,7 @@ export async function createStreamingManager<T extends CreateStreamOptions>(
                 }
 
                 try {
-                    if (state === ConnectionState.Connected) {
-                        await close(streamIdFromServer, session_id).catch(_ => {});
-                    }
+                    await close(streamIdFromServer, session_id);
```
nivzelber, line 244: Any reason we had the connection state check before?

nivzelber (changes requested):

src/createStreamingManager.ts
```diff
@@ -139,63 +139,71 @@ export async function createStreamingManager<T extends CreateStreamOptions>(
...
-                    {
-                        candidate: event.candidate.candidate,
-                        sdpMid: event.candidate.sdpMid,
-                        sdpMLineIndex: event.candidate.sdpMLineIndex,
-                    },
-                    session_id
-                );
-            } else {
-                addIceCandidate(streamIdFromServer, { candidate: null }, session_id);
+    const establishWebRtcConnection = async () => {
```
HaMMeRSI, line 142: move this function out of here

HaMMeRSI (approved):

PR #119, opened 2025-02-25
bugfix / chat silent error and retry when stream ttl edgecase

dani-did:


benyael15 (commented):
isnt the change to that type RequestOptions requires breaking changes?

src/createAgentManager.ts
```diff
@@ -93,6 +93,13 @@ async function newChat(
     }
 }
 
+class SdkError extends Error {
```
HaMMeRSI, line 96: move this to its own file

src/createAgentManager.ts
```diff
@@ -113,10 +121,24 @@ function initializeStreamAndChat(
...
+            const returnedChatMode: ChatMode = chat?.chat_mode || initialChatMode;
+
+            if (returnedChatMode !== initialChatMode) {
+                options.mode = returnedChatMode;
+                options.callbacks.onModeChange?.(returnedChatMode);
+
+                if (returnedChatMode === ChatMode.TextOnly) {
+                    options.callbacks?.onError?.(
+                        new SdkError({
+                            kind: 'ModeUnavailable',
```
HaMMeRSI, line 133: mode is present, the name does not say what it means, change it

HaMMeRSI (approved):

PR #125, opened 2025-03-06
Add check to video state, pc channel + stats

HaMMeRSI:
Add check to video state, pc channel + stats


osimhi213:
As discussed I need to stabilize the datachannel events

PR #129, opened 2025-04-03
feature: greet via the sdk

dor-eitan:
This PR removes the automatic greeting from the streamer's init operation. Instead, greetings are now explicitly handled via the SDK using the speak mechanism upon establishing a new connection.

### DIFF
--- a/src/services/chat/intial-messages.ts
+++ b/src/services/chat/intial-messages.ts
@@ -1,19 +1,32 @@
-import { Agent, Message } from '$/types';
+import { Agent, AgentManagerOptions, Message } from '$/types';
 import { getRandom } from '$/utils';
+import { getAgentType } from '$/utils/agent';
 
-export function getGreetings(agent: Agent) {
-    const greetings = agent.greetings?.filter(greeting => greeting.length > 0) ?? [];
-
-    if (greetings.length > 0) {
-        return greetings[Math.floor(Math.random() * greetings.length)];
-    }
-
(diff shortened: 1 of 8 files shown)


src/services/chat/intial-messages.ts
```diff
@@ -1,21 +1,34 @@
-import { Agent, Message } from '$/types';
+import { Agent, AgentManagerOptions, Message } from '$/types';
 import { getRandom } from '$/utils';
+import { getAgentType } from '$/utils/agent';
+
+function shouldGreet(agent: Agent, options: AgentManagerOptions): boolean {
+    return !!(getAgentType(agent.presenter) !== 'clip' && options.streamOptions?.streamGreeting);
```
benyael15, line 6: clips-v2 is supporting greeting - not only talks
dod u validate

src/services/streaming-manager/index.ts
```diff
@@ -122,8 +122,17 @@ export async function createStreamingManager<T extends CreateStreamOptions>(
     pcDataChannel.onopen = () => {
         isDatachannelOpen = true;
 
-        if ((!agent.stream_warmup && !agent.stream_greeting) || isConnected) {
```
benyael15, line 125: why is it the right place to trigger the speak? on data channel open?
pls validate with @osimhi213 and @HaMMeRSI

benyael15 (changes requested):
validate:
1. the point where we connected and trigger the speak(greeting)
2. validate functioning of talk && clips-v2
3. whats the user behvaiour of sdk user try to use it via the sdk against clips-v1-streaming?

PR #131, opened 2025-04-06
feature: charge credits for speak

dor-eitan:
passing to the speak method the chat id and the agent id to charge credits


osimhi213:
metadata is missing on SendClipStreamPayload.
Do we need to use this interface? why did we have it in the first place @HaMMeRSI ?

dor-eitan:
> metadata is missing on SendClipStreamPayload. Do we need to use this interface? why did we have it in the first place @HaMMeRSI ?

@osimhi213 according to the clips-controller, it does exist. the types here are probably wrong. 
This is how the current clips/talks of chats are getting its agent/chat id

nivzelber (approved):

PR #134, opened 2025-04-27
feature: fluent streaming

dor-eitan:
Task: https://app.asana.com/1/856614567666442/project/1209921650299098/task/1209210693250775?focus=true

- Added a `streamType` of `Leagacy`/`Fluent` that returned from the create stream
- Added a new callback of `onAgentStateChange` that reflects streaming/idle
- disabled warmup when for fluent

### DIFF
--- a/demo/hooks/useAgentManager.ts
+++ b/demo/hooks/useAgentManager.ts
@@ -1,13 +1,13 @@
 import { createAgentManager } from '$/services/agent-manager';
 import {
+    AgentActivityState,
     AgentManager,
-    AgentState,
     Auth,
     ChatMode,
     ConnectionState,
     Message,
-    StreamingState,
     StreamType,
+    StreamingState,
 } from '$/types';
 import { useCallback, useEffect, useState } from 'preact/hooks';
 
@@ -33,7 +33,7 @@ export function useAgentManager(props: UseAgentManagerOptions) {
     const [isSpeaking, setIsSpeaking] = useState(false);
     const [messages, setMessages] = useState<Message[]>([]);
     const [videoState, setVideoState] = useState(StreamingState.Stop);
-    const [agentState, setAgentState] = useState(AgentState.Idle);
+    const [agentActivityState, setAgentActivityState] = useState(AgentActivityState.Idle);
     const [srcObject, setSrcObject] = useState<MediaStream | null>(null);
     const [agentManager, setAgentManager] = useState<AgentManager | null>(null);
     const [connectionState, setConnectionState] = useState<ConnectionState>(ConnectionState.New);
@@ -41,9 +41,9 @@ export function useAgentManager(props: UseAgentManagerOptions) {
 
     useEffect(() => {
         if (streamType === StreamType.Fluent) {
-            setIsSpeaking(agentState === AgentState.Speaking);
+            setIsSpeaking(agentActivityState === AgentActivityState.Talking);
         }
-    }, [agentState, streamType]);
+    }, [agentActivityState, streamType]);
 
     useEffect(() => {
         if (streamType === StreamType.Legacy) {
@@ -70,7 +70,7 @@ export function useAgentManager(props: UseAgentManagerOptions) {
                         setVideoState(state);
                     },
                     onConnectivityStateChange(state) {
-                        console.log("onConnectivityStateChange: ", state);
+                        console.log('onConnectivityStateChange: ', state);
                     },
                     onNewMessage(newMessages, _type) {
                         setMessages([...newMessages]);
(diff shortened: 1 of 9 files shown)


src/services/streaming-manager/index.ts
```diff
@@ -40,23 +43,78 @@ function mapConnectionState(state: RTCIceConnectionState): ConnectionState {
...
+}: {
+    statsSignal?: StreamingState;
+    dataChannelSignal?: StreamingState;
+    onVideoStateChange: StreamingManagerOptions['callbacks']['onVideoStateChange'];
+    report?: VideoRTCStatsReport;
+}) {
     if (statsSignal === StreamingState.Start && dataChannelSignal === StreamingState.Start) {
         onVideoStateChange?.(StreamingState.Start);
-    } else if (statsSignal === StreamingState.Stop && dataChannelSignal === StreamingState.Stop) {
+        return;
```
nivzelber, line 59: I don't think you need to return here

src/services/streaming-manager/index.ts
```diff
@@ -40,23 +43,78 @@ function mapConnectionState(state: RTCIceConnectionState): ConnectionState {
...
+    onVideoStateChange,
+    onAgentStateChange,
+    report,
+}: {
+    statsSignal?: StreamingState;
+    dataChannelSignal?: StreamingState;
+    onVideoStateChange: StreamingManagerOptions['callbacks']['onVideoStateChange'];
+    onAgentStateChange?: StreamingManagerOptions['callbacks']['onAgentStateChange'];
+    report?: VideoRTCStatsReport;
+}) {
```
nivzelber, line 78: Nice!

src/types/stream/stream.ts
```diff
@@ -15,9 +15,19 @@ export enum StreamingState {
 export enum ConnectivityState {
     Strong = 'STRONG',
     Weak = 'WEAK',
-    Unknown = 'UNKNOWN'
+    Unknown = 'UNKNOWN',
 }
 
+export enum AgentState {
+    Idle = 'IDLE',
+    Speaking = 'SPEAKING',
```
nivzelber, line 23: We tend to call this part `TALKING`
It's already a convention around here

src/services/streaming-manager/index.ts
```diff
@@ -124,19 +191,23 @@ export async function createStreamingManager<T extends CreateStreamOptions>(
...
-        if (event.data === 'stream/started') {
-            dataChannelSignal = StreamingState.Start;
-        } else if (event.data === 'stream/done') {
-            dataChannelSignal = StreamingState.Stop;
-        }
+        if (event.data in DataChannelSignalMap) {
+            dataChannelSignal = DataChannelSignalMap[event.data];
 
-        handleStreamState(statsSignal, dataChannelSignal, callbacks.onVideoStateChange);
+            handleStreamState({
```
nivzelber, line 203: We use to **always** call this function. Don't we want to to it know also? Or at least warn here that we got an unsupported dataChannelSignal?

src/types/stream/stream.ts
```diff
@@ -15,9 +15,19 @@ export enum StreamingState {
 export enum ConnectivityState {
     Strong = 'STRONG',
     Weak = 'WEAK',
-    Unknown = 'UNKNOWN'
+    Unknown = 'UNKNOWN',
 }
 
+export enum AgentState {
```
nivzelber, line 21: This is a very confusing name. It can refer to a lot of stuff.
Think of something better.

demo/hooks/useAgentManager.ts
```diff
@@ -23,9 +32,24 @@ export function useAgentManager(props: UseAgentManagerOptions) {
...
     const [isSpeaking, setIsSpeaking] = useState(false);
     const [messages, setMessages] = useState<Message[]>([]);
+    const [videoState, setVideoState] = useState(StreamingState.Stop);
+    const [agentState, setAgentState] = useState(AgentState.Idle);
     const [srcObject, setSrcObject] = useState<MediaStream | null>(null);
     const [agentManager, setAgentManager] = useState<AgentManager | null>(null);
     const [connectionState, setConnectionState] = useState<ConnectionState>(ConnectionState.New);
+    const streamType = agentManager?.getStreamType();
+
+    useEffect(() => {
```
nivzelber, line 42: One of the simplest, most beautiful pieces of code we have in this repo. Good work.

nivzelber (changes requested):

src/services/streaming-manager/index.ts
dor-eitan: previously, indeed we called that function always. BUT, even if the dataChannelSignal wasn't supported, we just called it with the prev signal and not with the new one.
this doesn't make any sense to me, why to "change" the state to the same previous state?
regarding the warning, its an sdk to the customers and we already have events that we don't use. won't it just spam them without any reason?

nivzelber (approved):

PR #144, opened 2025-05-19
bugfix: speak when textual

dor-eitan:


src/services/agent-manager/index.ts
```diff
@@ -413,6 +413,16 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
                 options.callbacks.onNewMessage?.([...items.messages], 'answer');
             }
 
+            const isTextual = [ChatMode.TextOnly, ChatMode.Playground, ChatMode.Maintenance].includes(items.chatMode);
```
nivzelber, line 416: this can be a util method

src/services/agent-manager/index.ts
```diff
@@ -413,6 +413,16 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
...
+
+            // If the current chat is textual, we shouldn't activate the TTS.
+            if (items.chat && isTextual) {
+                return Promise.resolve({
```
nivzelber, line 420: change change the function to async if you must

nivzelber (changes requested):

src/services/agent-manager/index.ts
dor-eitan: ?

nivzelber (approved):

PR #152, opened 2025-06-25
Feature: interrupt

dor-eitan:


demo/app.tsx
```diff
@@ -81,6 +82,10 @@ export function App() {
                             Send to Chat
                         </button>
 
+                        <button onClick={interrupt} disabled={connectionState !== ConnectionState.Connected || !fluent}>
+                            Interrupt
```
nivzelber, line 86: Should interrupt be the *commercial* name of this feature?

src/services/streaming-manager/index.ts
```diff
@@ -292,6 +291,22 @@ export async function createStreamingManager<T extends CreateStreamOptions>(
                 clearInterval(videoStatsInterval);
             }
         },
+        /**
+         * Method to send data channel messages to the server
+         */
+        sendDataChannelMessage(payload: string) {
+            if (!isConnected || pcDataChannel.readyState !== 'open') {
+                log('Data channel is not ready for sending messages');
+                return;
```
nivzelber, line 300: I think we should indicate this to the caller. This looks like a good place to throw an error (and catch is someplace else)

src/services/agent-manager/index.ts
```diff
@@ -49,14 +51,41 @@ export interface AgentManagerItems {
...
+    let queuedInterrupt = false;
+    let currentVideoId: string | undefined;
+
+    const resetInterruptState = (): void => {
+        chatRequestPending = false;
+        queuedInterrupt = false;
+        currentVideoId = undefined;
+    };
+
+    const options: AgentManagerOptions = {
```
nivzelber, line 67: I think we have another convention for doing those sort of stuff - i.e. getting in the middle of a callback call.
I can't remember where but I remember we do have one. Can you check if I'm right? If so - update it.

nivzelber (approved):
Looks good. I have some sound comments, take a look.
Either way I saw you haven't chanegd the behavior with the queued interrupt. Nothing else came to mind?

src/services/agent-manager/index.ts
```diff
@@ -307,6 +349,8 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
 
                 return response;
             } catch (e) {
+                queuedInterrupt = false;
```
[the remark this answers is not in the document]
dor-eitan: if there is queued interrupt, i set it to false earlier. if for any reason something failed - then yes, clear it

src/services/agent-manager/index.ts
nivzelber: which new version?

nivzelber (approved):

demo/app.tsx
benyael15: yes :)

PR #170, opened 2025-07-30
video id from data channel

dariusz-did:
https://app.asana.com/1/856614567666442/project/1210781595684122/task/1210854541178832

https://github.com/de-id/agents-ui/pull/635

### DIFF
--- a/src/services/agent-manager/connect-to-manager.ts
+++ b/src/services/agent-manager/connect-to-manager.ts
@@ -127,7 +127,7 @@ function trackLegacyVideoAnalytics(
     }
 }
 
-type ContentToManagerOptions = AgentManagerOptions & {
+type ConnectToManagerOptions = AgentManagerOptions & {
     callbacks: AgentManagerOptions['callbacks'] & {
         onVideoIdChange?: (videoId: string | null) => void;
     };
@@ -135,7 +135,7 @@ type ContentToManagerOptions = AgentManagerOptions & {
 
 function connectToManager(
     agent: Agent,
-    options: ContentToManagerOptions,
+    options: ConnectToManagerOptions,
     analytics: Analytics
 ): Promise<StreamingManager<CreateStreamOptions>> {
     latencyTimestampTracker.reset();
@@ -191,7 +191,7 @@ function connectToManager(
 
 export async function initializeStreamAndChat(
     agent: Agent,
-    options: ContentToManagerOptions,
+    options: ConnectToManagerOptions,
     agentsApi: AgentsAPI,
     analytics: Analytics,
     chat?: Chat


src/services/agent-manager/index.ts
```diff
@@ -166,7 +181,7 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
     return {
         agent: agentEntity,
         getStreamType: () => items.streamingManager?.streamType,
-        getIsInterruptAvailable: () => items.streamingManager?.interruptAvailable ?? false,
+        getIsInterruptAvailable: () => (items.streamingManager?.interruptAvailable && !!videoId) ?? false,
```
dor-eitan, line 184: this should be if the feature is on/off
checking for video id should be in the validations

src/services/streaming-manager/index.ts
```diff
@@ -248,10 +248,13 @@ export async function createStreamingManager<T extends CreateStreamOptions>(
         [StreamEvents.StreamStarted]: handleStreamVideoEvent,
         [StreamEvents.StreamDone]: handleStreamVideoEvent,
         [StreamEvents.StreamReady]: handleStreamReadyEvent,
-    } satisfies Partial<{ [K in StreamEvents]: DataChannelMessageHandler<K> }>
+    } satisfies Partial<{ [K in StreamEvents]: DataChannelMessageHandler<K> }>;
 
     pcDataChannel.onmessage = (event: MessageEvent) => {
         const { subject, data } = parseDataChannelMessage(event.data);
+
+        callbacks.onMessage?.(subject, data);
```
dor-eitan, line 256: why?

src/types/entities/agents/manager.ts
```diff
@@ -96,6 +96,13 @@ interface ManagerCallbacks {
      * @param state - AgentActivityState
      */
     onAgentActivityStateChange?(state: AgentActivityState): void;
+
+    /**
+     * Optional callback function that will be triggered each time a message is received
+     * @param subject - StreamEvents
+     * @param data - string | object
+     */
+    onMessage?(subject: StreamEvents, data: string | object): void;
```
dor-eitan, line 105: why do we need to add a new callback to the user? whats its purpose? how they gonna use it?
also, if they pass it - you override it

dor-eitan (changes requested):

src/services/agent-manager/index.ts
```diff
@@ -106,7 +119,23 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
...
+                return initializeStreamAndChat(
+                    agentEntity,
+                    {
+                        ...options,
+                        callbacks: {
+                            ...options.callbacks,
+                            onMessage: (event: StreamEvents, data: string | object) => {
+                                updateVideoId(event, data);
+
+                                options.callbacks.onMessage?.(event, data);
```
dor-eitan, line 131: so you trigger that callback on every message in the data channel? not all messages are relevant to video id and in the future the gap will be even bigger.

dor-eitan: also, the onMessage callback usage is confusing. its a data channel message and not some general message. and also we are not injecting a user callback inside the data channel messages flow. i think it should be more managed.

src/services/streaming-manager/index.ts
dor-eitan, line 256: I would not inject here a user callback as-is. we might have many more data channel messages and this design is like the wild west. 
a callback should be attached to a specific action.
I think a better thing here will be to pass the callback of setting the video id ONLY to the stream start/end events

dor-eitan (approved):
yesssss

PR #172, opened 2025-08-04
mixpanel for templates

dariusz-did:
https://app.asana.com/1/856614567666442/project/1210781595684122/task/1210882333225121?focus=true

### DIFF
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -179,6 +179,9 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
                 chatId: items.chat?.id,
                 agentId: agentEntity.id,
                 mode: items.chatMode,
+                access: agentEntity.access,
+                name: agentEntity.preview_name,
+                ...(agentEntity.access === 'public' ? { from: 'agent-template' } : {}),
             });
         },
         async reconnect() {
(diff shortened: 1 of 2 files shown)


src/services/agent-manager/index.ts
```diff
@@ -179,6 +179,9 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
                 chatId: items.chat?.id,
                 agentId: agentEntity.id,
                 mode: items.chatMode,
+                access: agentEntity.access,
+                name: agentEntity.preview_name,
+                ...(agentEntity.access === 'public' ? { from: 'agent-template' } : {}),
```
dor-eitan, line 184: what about removing ALL of those props and just enrich the mixpanel after the agent init (I think we already have such flow) and put here only the gap. for example, track only the "event: connect"

src/services/chat/index.ts
```diff
@@ -20,9 +20,12 @@ export async function createChat(
 
             analytics.track('agent-chat', {
                 event: 'created',
-                chat_id: chat.id,
-                agent_id: agent.id,
+                chatId: chat.id,
+                agentId: agent.id,
```
dor-eitan, line 24: communicate it to PM/Ben so it won't break current dashboards

dor-eitan (approved):

PR #203, opened 2025-09-07
Feature/streams/v2

ariksfaradi:
## Add LiveKit Streaming Manager

This PR introduces a new LiveKit-based streaming manager alongside the existing WebRTC implementation, providing an alternative streaming technology for agent interactions.

### Key Features:
- **LiveKit Integration**: Implements real-time communication using LiveKit SDK for enhanced streaming capabilities
- **Shared Architecture**: Both WebRTC and LiveKit managers now use a common `StreamingManager` interface and shared utilities
- **Consistent API**: Maintains the same interface (`speak`, `disconnect`, `sendDataChannelMessage`) for seamless integration
- **Data Channel Support**: Handles streaming events (start/stop) and video ID changes via LiveKit's data channel
- **Error Handling**: Robust error handling and connection state management
- **Analytics Integration**: Tracks streaming events and enriches analytics data

### Technical Details:
- Added `livekit-client` dependency
- Created `common.ts` with shared utilities and types
- Refactored both managers to use shared logging and error handling
- Added `StreamType.LiveKit` enum value
- Factory pattern automatically selects the appropriate manager based on the agent type

### Backward Compatibility:
- No breaking changes to existing WebRTC functionality
- Same API surface for both streaming technologies

### DIFF
--- a/src/services/streaming-manager/factory.test.ts
+++ b/src/services/streaming-manager/factory.test.ts
@@ -1,0 +1,82 @@
+import { AgentFactory, StreamingManagerOptionsFactory } from '../../test-utils/factories';
+import { CreateStreamOptions, Providers, StreamingManagerOptions } from '../../types';
+import { createStreamingManager } from './factory';
+
+const mockCreateWebRTCStreamingManager = jest.fn();
+jest.mock('./webrtc-manager', () => ({
+    createWebRTCStreamingManager: (...args: any[]) => mockCreateWebRTCStreamingManager(...args),
+}));
+
+const mockCreateLiveKitStreamingManager = jest.fn();
+jest.mock('./livekit-manager', () => ({
+    createLiveKitStreamingManager: (...args: any[]) => mockCreateLiveKitStreamingManager(...args),
+}));
+
+describe('createStreamingManager', () => {
+    let mockStreamOptions: CreateStreamOptions;
+    let mockOptions: StreamingManagerOptions;
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+
+        mockStreamOptions = {
+            stream_warmup: false,
+        };
+
+        mockOptions = StreamingManagerOptionsFactory.build();
+    });
+
+    it('calls to createWebRTCStreamingManager when agent presenter type is talk', async () => {
+        const agent = AgentFactory.build({
+            presenter: {
+                type: 'talk',
+                source_url: 'https://example.com/presenter',
+                voice: {
+                    type: Providers.Microsoft,
+                    voice_id: 'voice-123',
+                },
+            },
+        });
+
+        await createStreamingManager(agent, mockStreamOptions, mockOptions);
+
+        expect(mockCreateWebRTCStreamingManager).toHaveBeenCalledWith(agent.id, mockStreamOptions, mockOptions);
+        expect(mockCreateLiveKitStreamingManager).not.toHaveBeenCalled();
+    });
+
+    it('calls to createWebRTCStreamingManager when agent presenter type is clip', async () => {
+        const agent = AgentFactory.build({
+            presenter: {
+                type: 'clip',
+                driver_id: 'driver-123',
+                presenter_id: 'presenter-123',
+                voice: {
+                    type: Providers.Microsoft,
+                    voice_id: 'voice-123',
+                },
+            },
+        });
+
+        await createStreamingManager(agent, mockStreamOptions, mockOptions);
+
+        expect(mockCreateWebRTCStreamingManager).toHaveBeenCalledWith(agent.id, mockStreamOptions, mockOptions);
+        expect(mockCreateLiveKitStreamingManager).not.toHaveBeenCalled();
+    });
+
+    it('calls to createLiveKitStreamingManager when agent presenter type is expressive', async () => {
+        const agent = AgentFactory.build({
+            presenter: {
+                type: 'expressive',
+                voice: {
+                    type: Providers.Microsoft,
+                    voice_id: 'voice-123',
+                },
+            },
+        });
+
+        await createStreamingManager(agent, mockStreamOptions, mockOptions);
+
+        expect(mockCreateLiveKitStreamingManager).toHaveBeenCalledWith(agent.id, mockStreamOptions, mockOptions);
+        expect(mockCreateWebRTCStreamingManager).not.toHaveBeenCalled();
+    });
+});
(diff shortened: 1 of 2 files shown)


src/services/streaming-manager/common.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Shared type for all streaming managers (LiveKit, WebRTC, etc.)
+ * This type represents the return value of any streaming manager implementation
+ */
+export type StreamingManager<T extends CreateStreamOptions> = {
+    /**
+     * Method to send request to server to get clip or talk depending on payload
+     * @param payload The payload to send to the streaming service
+     */
+    speak(payload: any): Promise<any>;
```
dor-eitan, line 15: can we avoid any?

src/services/streaming-manager/common.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Id of current streaming session
+     */
+    streamId: string;
+
+    /**
+     * Type of streaming implementation being used
+     */
+    streamType: any;
```
dor-eitan, line 41: StreamType enum?

src/types/stream/stream.ts
```diff
@@ -50,6 +50,7 @@ export enum ConnectionState {
 export enum StreamType {
     Legacy = 'legacy',
     Fluent = 'fluent',
+    LiveKit = 'livekit',
```
dor-eitan, line 53: Im not sure that liveKit should be in the same type with fluent/legacy. 
fluent/legacy = webrtc and i think that in livekit it will also be fluent?
the ui is also using that stream type field.
maybe connection type?

package.json
```diff
@@ -31,6 +31,9 @@
         "test-build": "node .infra/build.js -m development",
         "build:docs": "typedoc"
     },
+    "dependencies": {
+        "livekit-client": "^2.0.0"
+    },
```
dor-eitan, line 36: I understand that we need it but our current sdk size is: 144 kB and the livekit-client sdk is 6.81 MB.
we try to keep it lean in terms of dependencies so it will be lightweight. 
not sure that there is much to do...

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+        log(`Track subscribed: ${track.kind} from ${participant.identity}`);
+
+        if (track.kind === 'video') {
+            const videoTrack = track as RemoteVideoTrack;
+            const mediaStream = new MediaStream([videoTrack.mediaStreamTrack]);
+            callbacks.onSrcObjectReady?.(mediaStream);
+        } else if (track.kind === 'audio') {
+            const audioTrack = track as RemoteAudioTrack;
+            const mediaStream = new MediaStream([audioTrack.mediaStreamTrack]);
+        }
```
dor-eitan, line 59: is it a case we want to handle? what are we doing with that audio mediaStream?

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+        log('Data received:', message);
+
+        // Handle data channel messages similar to WebRTC
+        try {
+            const data = JSON.parse(message);
+            if (data.subject === 'stream_started' && data.metadata?.videoId) {
+                videoId = data.metadata.videoId;
+                callbacks.onVideoIdChange?.(videoId);
+                callbacks.onVideoStateChange?.(StreamingState.Start);
+                callbacks.onAgentActivityStateChange?.(AgentActivityState.Talking);
```
dor-eitan, line 73: it depends on the stream type.
in "legacy":
- everything was managed via video state: start/stop. if its start then its talking.
in "fluent":
- we always have a video so once the stream started we change the video state to start
- when the avatar speak/stop - we change the agentActivityState

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+        'stream-type': StreamType.LiveKit,
+    });
+
+    return {
+        speak(payload: PayloadType<T>) {
+            if (!isConnected || !room) {
+                throw new Error('Room is not connected');
+            }
+
+            // Send speak request via data channel
```
dor-eitan, line 105: why not using your method: `sendDataChannelMessage`

dor-eitan: and speak via datachannel will be only in livekit implementation?

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -0,0 +1,154 @@
...
+            });
+        },
+
+        async disconnect() {
+            if (room) {
+                await room.disconnect();
+                room = null;
+            }
+            isConnected = false;
+            callbacks.onAgentActivityStateChange?.(AgentActivityState.Idle);
```
dor-eitan, line 127: will also need video state = stop or maybe only videostate=stop. 
need to validate the prev logic

src/services/streaming-manager/webrtc-manager.ts
```diff
@@ -0,0 +1,385 @@
...
+        case 'closed':
+            return ConnectionState.Closed;
+        case 'disconnected':
+            return ConnectionState.Disconnected;
+        case 'completed':
+            return ConnectionState.Completed;
+        default:
+            return ConnectionState.New;
+    }
+}
```
dor-eitan, line 45: I didn't see most of them in the livekit implementation. I believe we will need most of them such as fail/connecting etc

dor-eitan (commented):
the e2e in that pipeline is prod of agents-ui with the current build of the sdk. so it verifys the backward compatibility.
looks good! i do think we will need to be more accurate with the internal states

package.json
ariksfaradi: We will lose the livekit types - will need to convert to any or define in a local file

dor-eitan: do they export types package?

ariksfaradi: No, it's coupled in the client SDK

dor-eitan (approved):
lets verify that the e2e pass before pushing it
also, consider lazy load it as most of the users will not use it for now and it will dramatically increase the size of the package/ui snippet loading time

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -0,0 +1,196 @@
...
+                break;
+        }
+    });
+
+    // Handle connection errors
+    room.on(RoomEvent.ConnectionQualityChanged, (quality, participant) => {
+        if (participant?.isLocal && quality === 'poor') {
+            log('Connection quality is poor');
+        }
+    });
```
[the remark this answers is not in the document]
ariksfaradi: Maybe a new onConnectionQualityChanged cb?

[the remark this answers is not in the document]
dor-eitan: the current usage is:
`onConnectivityStateChange(ConnectivityState.Weak)`

PR #208, opened 2025-09-18
allow to pass mixpanel properties

dariusz-did:
https://app.asana.com/1/856614567666442/project/1211131497210590/task/1211245261664857?focus=true


src/services/analytics/mixpanel.ts
```diff
@@ -43,13 +44,17 @@ export function initializeAnalytics(config: AnalyticsOptions): Analytics {
...
+            id: config.distinctId,
+            ...(config.mixpanelAdditionalProperties || {}),
+        },
         isEnabled: config.isEnabled ?? true,
         getRandom,
         enrich(props: Record<string, any>) {
             this.additionalProperties = { ...this.additionalProperties, ...props };
         },
         async track(event: string, props?: Record<string, any>) {
+            console.log('track', event, props);
```
dor-eitan, line 57: remove

dor-eitan (commented):
The pr looks ok, but why do we actually need it?
what about just doing mixpanel.enrich in the ui?

dor-eitan (approved):

PR #362, opened 2026-04-18
fix: preserve multiple assistant messages within the same turn

dariusz-did:
## Summary

- Detects when a `chat/answer` event introduces a **new** assistant message (different `id`) vs. updates the current one (same/missing `id`), and appends instead of overwriting.
- Clears the streaming buffer when pushing the new message so its content does not inherit the previous message's partials.
- Backward compatible: answers with the same `id` or without `id` keep the existing overwrite behaviour.

## Problem

With client tools, the agent can emit more than one assistant message per turn:

```
user:      please book a meeting
assistant: ok, i'll book a meeting
[tool_call / tool_result]
assistant: i booked a meeting for you
```

The same also happens when the agent simply decides to send two assistant messages in a row without any tool call in between.

Today the SDK keeps only the last assistant message. In `processChatEvent`, when the last item in `items.messages` has role `assistant`, the code reuses it (`currentMessage = lastMessage`) and overwrites its `content`. The decision is based purely on role — the message `id` is never compared — so the first assistant message is silently replaced.

## Change

`src/services/socket-manager/message-queue.ts`: add an `isNewAssistantMessage` discriminator that fires only on `ChatProgress.Answer` when `data.id` is present and differs from the last assistant message's id. In that branch we `clearQueue()` (so `getMessageContent()` does not concatenate the previous message's answer into the new one) and push a fresh `Message`. `processChatEvent` now takes `clearQueue` as a parameter, wired from the closure in `createMessageEventQueue`.

## Caveats / follow-ups

- Assumes the backend emits distinct `id`s on `chat/answer` for distinct assistant messages. If the same `id` is sent twice, behaviour is unchanged (still overwrites) — which matches streaming partials semantics, so this is the right fallback. **Needs live verification** with a client-tool-enabled agent before merging out of draft.
- When the backend omits `id` entirely (legacy), we keep overwriting the last message. A covered test pins this behaviour.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npx jest --no-coverage` — 338/338 pass, includes 6 new cases under `multi-message assistant turns`:
  - append on distinct id
  - preserve first message across tool-call gap
  - overwrite on matching id
  - overwrite when id is missing (legacy)
  - no content leakage between messages
  - partial streaming still updates a single message
- [ ] Manual: run agent with a client tool, confirm both assistant bubbles render in the chat UI
- [ ] Manual: run agent that emits two messages without a tool call, confirm both render

## Related

- Asana: https://app.asana.com/1/856614567666442/project/1213882276505882/task/1214081847771983
- Touches the same area as #361 (`message-queue.ts`); expect a small merge conflict to resolve once either lands.

### DIFF
--- a/src/services/socket-manager/message-queue.ts
+++ b/src/services/socket-manager/message-queue.ts
@@ -73,10 +73,7 @@ function processChatEvent(
     // assistant messages in a row) is signalled by an answer event whose `id` differs from the
     // last assistant message. Without this branch, the SDK overwrites the previous message.
     const isNewAssistantMessage =
-        event === ChatProgress.Answer &&
-        data.id &&
-        lastMessage?.role === 'assistant' &&
-        lastMessage.id !== data.id;
+        event === ChatProgress.Answer && data.id && lastMessage?.role === 'assistant' && lastMessage.id !== data.id;
 
     let currentMessage: Message;
     if (lastMessage?.transcribed && lastMessage.role === 'user') {
--- a/src/services/socket-manager/message-queue.ts
+++ b/src/services/socket-manager/message-queue.ts
@@ -55,7 +55,8 @@ function processChatEvent(
     data: any,
     chatEventQueue: ChatEventQueue,
     items: AgentManagerItems,
-    onNewMessage: AgentManagerOptions['callbacks']['onNewMessage']
+    onNewMessage: AgentManagerOptions['callbacks']['onNewMessage'],
+    clearQueue: () => void
 ) {
     if (event === ChatProgress.Transcribe && data.content) {
         handleAudioTranscribedMessage(data, items, onNewMessage);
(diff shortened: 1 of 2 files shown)


src/services/socket-manager/message-queue.ts
```diff
@@ -68,16 +69,31 @@ function processChatEvent(
 
     const lastMessage = items.messages[items.messages.length - 1];
 
+    // A new assistant message within the same turn (e.g. after a client tool call, or several
+    // assistant messages in a row) is signalled by an answer event whose `id` differs from the
+    // last assistant message. Without this branch, the SDK overwrites the previous message.
+    const isNewAssistantMessage =
+        event === ChatProgress.Answer && data.id && lastMessage?.role === 'assistant' && lastMessage.id !== data.id;
```
dor-eitan: `ChatProgress.Answer` is the final answer. 
each message flow is:
`ChatProgress.Partial`
`ChatProgress.Partial`
`ChatProgress.Partial`
`ChatProgress.Answer`

so im not sure we should only check for Answer here. The "new" message will probably start with Partial

src/services/socket-manager/message-queue.ts
```diff
@@ -68,16 +69,31 @@ function processChatEvent(
...
             role: data.role || 'assistant',
             content: data.content || '',
             created_at: data.created_at || new Date().toISOString(),
         };
         items.messages.push(currentMessage);
+    } else if (isNewAssistantMessage) {
+        // Reset the streaming buffer so the next message does not inherit the previous one's content.
+        clearQueue();
+        currentMessage = {
+            id: data.id,
```
dor-eitan: id: data.id || `assistant-${Date.now()}`,

src/services/socket-manager/message-queue.ts
dariusz-did: Good catch — you are right, the new message starts with `Partial` events, so gating on `Answer` only meant the first partial of a new id would silently overwrite the previous message. Dropped the `event === Answer` check so the discriminator fires on both `Partial` and `Answer`, and added a test (`should append a new assistant message when partials arrive with a different id`) covering the full `Partial → Partial → Answer` flow for a second message. Fixed in f1a5515.

src/services/socket-manager/message-queue.ts
dariusz-did: Done — aligned with the rest of the file for consistency. Fixed in f1a5515.

dor-eitan (approved):
🚀

PR #371, opened 2026-04-28
feat(livekit): add replaceMicrophoneTrack for seamless mic device swap

dor-eitan:
## Summary
- Adds `replaceMicrophoneTrack(track: MediaStreamTrack)` on the LiveKit streaming manager. Calls `LocalAudioTrack.replaceTrack` under the hood so the publication (SSRC, trackSid) is preserved across mic device swaps — no `TrackUnpublished`/`TrackPublished` round-trip, no STT/VAD reset on the server.
- Surfaces the method through `StreamingManager` (optional, LiveKit-only) and the public `AgentManager` interface, with a `'not available'` error on the WebRTC path.
- Throws `'No microphone publication to replace'` when there is no live publication so callers can transparently fall back to `publishMicrophoneStream`. Also rejects when a publish is in flight or when the supplied track is not audio.
- PR 2 of the 4-part mic/audio plan. Adopted in agents-ui in a follow-up PR that bumps `@d-id/client-sdk`.

## Asana
https://app.asana.com/1/856614567666442/project/1214036289203665/task/1214079020472516

## Test plan
- [x] `yarn ci:test` — 373 tests pass, including 6 new `replaceMicrophoneTrack` cases in `livekit-manager.test.ts` (no-connection / no-publication / non-audio track / publish-in-progress / success no unpublish-publish / disconnect clears publication) and 2 pass-through cases in `agent-manager/index.test.ts`.
- [x] `yarn build` — clean. `replaceMicrophoneTrack` is present in `dist/src/types/entities/agents/manager.d.ts` and `dist/src/services/streaming-manager/common.d.ts`.
- [ ] Manual verification deferred to the agents-ui PR (requires a LiveKit-backed agent end-to-end).


src/services/agent-manager/index.ts
```diff
@@ -325,6 +325,12 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
             }
             return items.streamingManager.unpublishMicrophoneStream();
         },
+        async replaceMicrophoneTrack(track: MediaStreamTrack) {
```
ariksfaradi: async without await?
maybe instead `replaceMicrophoneTrack(track: MediaStreamTrack): Promise`?

src/services/streaming-manager/common.ts
```diff
@@ -47,6 +47,17 @@ export type StreamingManager<T extends CreateStreamOptions | CreateSessionV2Opti
...
     unpublishMicrophoneStream?(): Promise<void>;
 
+    /**
+     * Replace the currently published microphone MediaStreamTrack without
+     * unpublishing. Preserves the LiveKit publication (SSRC, trackSid) so the
+     * server sees continuous audio. Resolves once LiveKit has switched the
+     * underlying RTCRtpSender's track. Rejects if no publication exists —
+     * caller should fall back to `publishMicrophoneStream` in that case.
+     *
+     * Only implemented by the LiveKit streaming manager.
```
ariksfaradi: * supported only for livekit manager

src/services/streaming-manager/livekit-manager.test.ts
```diff
@@ -442,6 +444,131 @@ describe('LiveKit Streaming Manager - Microphone Stream', () => {
...
+
+            const videoTrack = createMockCameraTrack();
+
+            await expect(manager.replaceMicrophoneTrack?.(videoTrack)).rejects.toThrow(
+                'Microphone track must be an audio track'
+            );
+            expect(mockPublication.track.replaceTrack).not.toHaveBeenCalled();
+        });
+
+        it('should throw when a publish is already in progress', async () => {
```
ariksfaradi, line 482: can it breaks user session?
maybe we want to do nothing when publish is in progress?

ariksfaradi (approved):

src/services/streaming-manager/livekit-manager.test.ts
dor-eitan: Throwing here is intentional and the safer behavior, even though it looks aggressive. The agents-ui consumer wraps `replaceMicrophoneTrack` in a try/catch (`replaceTrackInSDK` in `useStreamPublishing.ts`): on throw it returns `false`, and `handleMicStream` falls back to the existing unpublish + publish path. The user gets the new device on the next swap with at most a single STT-pipeline bump — same as the pre-PR behavior.

If we silently no-op'd instead, `replaceTrackInSDK` would resolve with `undefined` (treated as success), set `publishedTrackIdRef = newTrack.id`, and return `true`. `handleMicStream` then skips both unpublish and publish. Result: SDK keeps the OLD track published while the UI thinks the new track is live — silent state corruption, much worse than throw + fallback.

In practice this branch is extremely unlikely to fire: `handleMicStream` only calls `replaceMicrophoneTrack` after `isMicrophonePublishedRef` is true, and the only way `microphoneState.isPublishing` is true at that exact moment is a concurrent in-flight `publishMicrophoneStream` (which the agents-ui side already serializes via its own `isPublishingRef`).

PR #376, opened 2026-04-28
fix: skip text interrupts on expressive agents to avoid races

ariksfaradi:
### Pull Request Type

🐛 BugFix

### Description

- Expressive (V2) streams now skip `interrupt({ type: 'text' })` — typing in the chat box while the agent is talking no longer fires redundant interrupts that race with prior unsettled ones.
- Refactor: each `StreamingManager` (livekit / webrtc) owns its own `interrupt(type)` method. `agent-manager` just forwards. WebRTC validation + payload preserved verbatim from the deleted `services/interrupt` helpers.

### Reference Links

- [Asana](https://app.asana.com/1/856614567666442/project/1213796129611649/task/1214338214296089?focus=true)


src/services/streaming-manager/webrtc-manager.ts
```diff
@@ -397,6 +400,25 @@ export async function createWebRTCStreamingManager<T extends CreateStreamOptions
         streamType,
         interruptAvailable: interruptAvailable ?? false,
         isInterruptible: true,
+
+        interrupt(_type: Interrupt['type']) {
+            if (!interruptAvailable) {
+                throw new Error('Interrupt is not enabled for this stream');
+            }
```
dor-eitan, line 407: Dont we need to validate this one also in livekit manager?

dor-eitan (approved):

PR #383, opened 2026-05-04
feat: consume interruptible flag from tool-call/started payload

dariusz-did:
### Pull Request Type

🔮 Feature

### Description

- SDK now consumes the `interruptible` field from `tool-call/started` payload, updates `currentInterruptible`, and emits `onInterruptibleChange` from `handleToolEvents`.
- Without this, UI cannot block interrupt during non-interruptible tools — `stream-video` events do NOT fire during tool execution, so the only signal is the tool-call payload itself.
- Backend already sends `interruptible: false` for all tools today (orchestrator commit b3b53259, deployed to prod).
- Flag carries over through `tool-call/done`/`error` until the next event flips it.
- Adds \`interruptible: boolean\` to \`ToolCallStartedPayload\` type.
- 3 new unit tests in \`livekit-manager.test.ts\` (63/63 passing).

### Reference Links

- [Asana](https://app.asana.com/1/856614567666442/project/1214036289203665/task/1214418722200514)

### DIFF
--- a/src/services/streaming-manager/livekit-manager.ts
+++ b/src/services/streaming-manager/livekit-manager.ts
@@ -364,9 +364,12 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
      */
     function handleToolEvents(subject: string, data: any): void {
         if (subject === StreamEvents.ToolCallStarted) {
+            const payload = data as ToolCallStartedPayload;
+            currentInterruptible = payload.interruptible !== false;
+            callbacks.onInterruptibleChange?.(currentInterruptible);
             currentActivityState = AgentActivityState.ToolActive;
             callbacks.onAgentActivityStateChange?.(AgentActivityState.ToolActive);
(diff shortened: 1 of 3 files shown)


src/services/streaming-manager/livekit-manager.ts
```diff
@@ -364,9 +364,12 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
      */
     function handleToolEvents(subject: string, data: any): void {
         if (subject === StreamEvents.ToolCallStarted) {
+            const payload = data as ToolCallStartedPayload;
+            currentInterruptible = payload.interruptible !== false;
```
dor-eitan, line 372: we will need to address it in the future if you have one tool that is interruptible and one that is isnt in parallel.  we will have race condition. 
for now it works.
lets add comment

dor-eitan (approved):

PR #393, opened 2026-05-08
fix: split multi-message assistant turns without relying on backend id

dariusz-did:
## Background — how the SDK builds an assistant message

The backend streams an assistant reply over **two kinds of chat events** delivered through the LiveKit data channel:

- **`chat/partial`** — one chunk of the message in flight. Each chunk has a `sequence` index so the SDK can re-order them as they arrive.
- **`chat/answer`** — the full final string for the message, sent once when the assistant finishes one logical reply.

The SDK keeps two structures inside `processChatEvent` (`socket-manager/message-queue.ts`):

```js
items.messages = [   // what the UI eventually renders
  { id: 'assistant-…', role: 'assistant', content: "I'll check the weather", parts: [...] }
]

chatEventQueue = {   // buffer used to assemble content from partials
  0: "I'll", 1: " check", 2: " the", 3: " weather"
}
```

`getMessageContent(chatEventQueue)` concatenates the buffer by sequence index. When the assembled string changes (or an `Answer` arrives) the SDK updates the last assistant message and notifies the UI via `onNewMessage`.

The whole routing decision boils down to one boolean: **is this event continuing the same logical message, or is it a new one?**

## The problem

The orchestrator no longer publishes a message id on `chat/partial` or `chat/answer` payloads, so the previous discriminator —

```js
const isNewAssistantMessage =
  data.id && lastMessage?.role === 'assistant' && lastMessage.id !== data.id;
```

— is permanently \`false\`, because \`data.id\` is always \`undefined\`. Three symptoms followed:

### Bug A — pre-tool ack + post-tool reply merged into one bubble

When an agent emits two logical assistant messages in the same user turn (typically a pre-tool acknowledgement, then the tool runs, then a post-tool reply), the SDK has no signal that the second stream is a new message. The post-tool partials are appended to the pre-tool message and rewrite the buffer slots starting at \`sequence: 0\`, so the UI either shows a single merged bubble or loses the pre-tool ack entirely.

### Bug B — clips/talks consecutive answers merge as well

Legacy flows that emit several \`chat/answer\` events in a row (no partials) hit the same fall-through path — every answer overwrites the previous one inside the same UI entry.

### Bug C — chat endpoint duplicates the streamed reply or inserts an empty bubble

\`manager.chat(userMessage)\` in \`agent-manager/index.ts\` pushes a user message, calls the REST chat endpoint, and then pushes an assistant message built from \`response.result\`. In streaming modes the actual reply is already being delivered over \`chat/partial\` + \`chat/answer\`, so this extra push produces one of two artefacts:

- **Empty placeholder bubble** when the chat endpoint returns \`response.result === ''\` (Fluent / voice modes).
- **Verbatim duplicate** when the chat endpoint echoes the assistant content that was just streamed (the symptom appeared after the second turn in our repro — the post-tool weather reply rendered twice).

The old code happened to mask both artefacts: every stream event was appended to the most recent assistant entry (the one pushed by \`chat()\`), so even a stale duplicate or an empty placeholder was overwritten in place. With proper message-boundary detection re-enabled they become visible.

### 1. Use \`chat/answer\` as the logical-message closer

\`chat/answer\` always closes one logical assistant message. The next \`chat/partial\` or \`chat/answer\` therefore opens a new one. Replace the id-mismatch check with a chat-event-boundary check, threaded through \`processChatEvent\` as an explicit parameter:

\`\`\`js
const isNewAssistantMessage =
  lastChatEventType === 'answer' && lastMessage?.role === 'assistant';
\`\`\`

\`lastMessageType\` already lives in the \`createMessageEventQueue\` closure; we only had to (a) pass it into \`processChatEvent\` and (b) update it directly when a chat event arrives in the \`onMessage\` wrapper (the previous update path ran through \`onNewMessage\`, which doesn't fire on empty-content partials, so the next partial would still see \`lastMessageType === 'answer'\` and open a second message).

### 2. Initial \`lastMessageType = 'user'\`

The original initial value was \`'answer'\`. That broke the locally-pushed greeting from \`agent.speak()\` because the first backend partial would see \`lastChatEventType === 'answer'\` and treat the greeting as a finished message instead of letting the stream overwrite the local push. Initial \`'user'\` keeps the original adopt-the-greeting behaviour intact.

### 3. Clear \`chatEventQueue\` after every \`chat/answer\`

The \`isMidAssistantStream\` proxy needs the buffer to be empty between turns; this also prevents stale partial slots from leaking into the next message.

### 4. \`clearQueue\` mutates the queue object

Previously \`clearQueue = () => (chatEventQueue = {})\` reassigned the outer-scope variable. The parameter inside \`processChatEvent\` still held the old reference, so the captured queue was never observed as cleared. Mutating the same object via \`delete\` propagates the reset to every captured reference.

### 5. Populate \`parts\` at message-creation time

Pushing a message with \`parts: []\` and updating later caused the UI to briefly observe a parts-empty assistant entry (which the existing \`isLoading && isLastAssistantMessage\` rule then hid). Computing \`parseMessagePartsMemo(initialContent)\` at push time eliminates the empty-parts window so no visible message is ever momentarily hidden.

### 6. Skip the \`chat()\` push when the stream already delivered the assistant message

\`agent-manager/index.ts\` now checks whether the most recent message after \`sendChatRequest()\` is already an assistant entry (which means the streaming pipeline produced a reply during this turn). If it is, we skip both the \`items.messages.push\` and the \`onNewMessage('answer')\` notification. Textual-only modes — where no stream arrives — still hit the original fallback because the last entry remains the just-pushed user message.

This is the minimal change that fixes Bug C without coupling the chat endpoint to a specific \`chatMode\` or streaming-manager type.

## Compatibility

- **Fluent (V2 / LiveKit):** post-tool reply renders as a second bubble alongside the pre-tool ack, no merged content, no duplicate.
- **Clips / talks (V1):** consecutive \`chat/answer\` events open new messages because \`lastChatEventType === 'answer'\` resolves correctly for every transition.
- **Textual modes (no stream):** unchanged — \`chat()\` still inserts the assistant message because the stream never produces one.
- **Greeting via \`agent.speak()\`:** unchanged — initial \`lastMessageType === 'user'\` lets the first backend partial overwrite the locally-pushed entry exactly as before.

## Test plan

- [x] Live test: greeting renders once; pre-tool ack and post-tool reply both stream as separate bubbles during the same turn; no duplicate, no empty placeholder.
- [x] \`yarn type-check\` — clean.
- [x] \`yarn lint\` — clean (Prettier).

### DIFF
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -447,15 +447,22 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
                 const chatId = await initializeChat();
                 const response = await sendChatRequest([...items.messages], chatId);
 
-                items.messages.push({
-                    id: getRandom(),
-                    role: 'assistant',
-                    content: response.result || '',
-                    parts: parseMessagePartsMemo(response.result || ''),
-                    created_at: new Date().toISOString(),
-                    context: response.context,
-                    matches: response.matches,
-                });
+                // Skip the assistant push entirely when `response.result` is empty — streaming
+                // modes deliver the actual reply through `chat/partial` + `chat/answer`, so the
+                // REST endpoint commonly returns an empty string that would otherwise leave a
+                // ghost assistant entry in `items.messages` (a "" bubble hidden by CSS but still
+                // exported in transcripts).
+                if (response.result) {
+                    items.messages.push({
+                        id: getRandom(),
+                        role: 'assistant',
+                        content: response.result,
+                        parts: parseMessagePartsMemo(response.result),
+                        created_at: new Date().toISOString(),
+                        context: response.context,
+                        matches: response.matches,
+                    });
+                }
 
                 analytics.track('agent-message-send', {
                     event: 'success',
--- a/src/services/socket-manager/message-queue.ts
+++ b/src/services/socket-manager/message-queue.ts
@@ -81,7 +81,7 @@ function processChatEvent(
     let currentMessage: Message;
     if (lastMessage?.role === 'assistant' && !isNewAssistantMessage) {
         currentMessage = lastMessage;
-    } else if (!lastMessage || (lastMessage.transcribed && lastMessage.role === 'user') || isNewAssistantMessage) {
+    } else if (!lastMessage || lastMessage.role === 'user' || isNewAssistantMessage) {
         if (isNewAssistantMessage) {
             // Reset the streaming buffer so the next message does not inherit the previous one's content.
             clearQueue();
@@ -174,8 +174,6 @@ export function createMessageEventQueue(
                     // Clear the streaming buffer so the next turn's content does not inherit
                     // leftover partial slots from this one.
                     clearQueue();
-                } else {
-                    lastAssistantMessageType = null;
                 }
 
                 if (chatEvent === ChatProgress.Answer) {
--- a/src/services/socket-manager/message-queue.ts
+++ b/src/services/socket-manager/message-queue.ts
@@ -58,7 +58,8 @@ function processChatEvent(
     chatEventQueue: ChatEventQueue,
     items: AgentManagerItems,
     onNewMessage: AgentManagerOptions['callbacks']['onNewMessage'],
-    clearQueue: () => void
+    clearQueue: () => void,
+    lastAssistantMessageType: 'partial' | 'answer' | null
 ) {
     if (event === ChatProgress.Transcribe && data.content) {
         handleAudioTranscribedMessage(data, items, onNewMessage);
@@ -71,12 +72,11 @@ function processChatEvent(
 
     const lastMessage = items.messages[items.messages.length - 1];
 
-    // A new assistant message within the same turn (e.g. after a client tool call, or several
-    // assistant messages in a row) is signalled by a chat event whose `id` differs from the
-    // last assistant message. The new message typically starts with `Partial` events and ends
-    // with `Answer`, so both branches must detect the id change — otherwise the SDK overwrites
-    // the previous message on the first partial of the new one.
-    const isNewAssistantMessage = data.id && lastMessage?.role === 'assistant' && lastMessage.id !== data.id;
+    // `chat/answer` closes a logical assistant message: the next chat event (Partial or Answer)
+    // starts a new one. This covers post-tool replies that follow a pre-tool ack within the same
+    // turn, and consecutive answers in clips/talks flows. The orchestrator does not send a
+    // message id, so the previous id-mismatch heuristic does not apply.
+    const isNewAssistantMessage = lastAssistantMessageType === 'answer';
 
     let currentMessage: Message;
     if (lastMessage?.role === 'assistant' && !isNewAssistantMessage) {
@@ -123,14 +123,24 @@ export function createMessageEventQueue(
     agentEntity: Agent,
     onStreamDone: () => void
 ) {
-    let chatEventQueue: ChatEventQueue = {};
-    const clearQueue = () => (chatEventQueue = {});
-    let lastMessageType: 'answer' | 'partial' | 'user' = 'answer';
+    const chatEventQueue: ChatEventQueue = {};
+    const clearQueue = () => {
+        // Mutate the queue object so closures that captured it (e.g. the parameter inside
+        // `processChatEvent`) observe the reset — reassigning the outer-scope variable would
+        // leave those captured references pointing at the old, populated object.
+        for (const key of Object.keys(chatEventQueue)) {
+            delete chatEventQueue[key as keyof ChatEventQueue];
+        }
+    };
+    // Tracks the last assistant chat event in the current turn. `null` means "no assistant stream
+    // is in progress" (initial state, or after a non-assistant event). The flag stays `null` for
+    // the first backend partial of the greeting so it adopts the locally-pushed `agent.speak()`
+    // entry instead of opening a new message.
+    let lastAssistantMessageType: 'partial' | 'answer' | null = null;
     const onNewMessage: AgentManagerOptions['callbacks']['onNewMessage'] = (messages, event) => {
         if (event === 'user') {
             clearQueue();
         }
-        lastMessageType = event;
         options.callbacks.onNewMessage?.(messages, event);
     };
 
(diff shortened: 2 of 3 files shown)


src/services/agent-manager/index.ts
```diff
@@ -447,22 +447,34 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
...
-                    parts: parseMessagePartsMemo(response.result || ''),
-                    created_at: new Date().toISOString(),
-                    context: response.context,
-                    matches: response.matches,
-                });
+                // Stream-based modes (Fluent / V1 voice / etc.) deliver the assistant reply via
+                // `chat/partial` + `chat/answer` events through `processChatEvent` — pushing
+                // `response.result` again would create an empty placeholder when the stream is
+                // still in flight, or a content duplicate when the chat endpoint echoes the
+                // already-streamed message. Only push when the stream produced nothing this turn
```
dor-eitan: if "Only push when the stream produced nothing this turn" -> why the condition isnt `!!response.result`

src/services/agent-manager/index.ts
```diff
@@ -447,22 +447,34 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
...
+                // (e.g. textual modes without a streaming manager) so the chat-endpoint result
+                // remains a valid fallback.
+                const lastAfterRequest = items.messages[items.messages.length - 1];
+                const streamDeliveredThisTurn = lastAfterRequest?.role === 'assistant';
+
+                if (!streamDeliveredThisTurn) {
```
dor-eitan: what do you really want to check here? that the items.messages includes something after we requested for new message? If so, does the next code will work/easier?
on line 437, save the id as const: `const messageId = getRandom()`
then you can check here if the last message id is the id we assigned

dor-eitan: also, are you sure it even changes something? because we only send the callback later if response.result (line 477)

src/services/socket-manager/message-queue.ts
```diff
@@ -71,12 +72,11 @@ function processChatEvent(
...
-    // assistant messages in a row) is signalled by a chat event whose `id` differs from the
-    // last assistant message. The new message typically starts with `Partial` events and ends
-    // with `Answer`, so both branches must detect the id change — otherwise the SDK overwrites
-    // the previous message on the first partial of the new one.
-    const isNewAssistantMessage = data.id && lastMessage?.role === 'assistant' && lastMessage.id !== data.id;
+    // `chat/answer` closes a logical assistant message: the next chat event (Partial or Answer)
+    // starts a new one. This covers post-tool replies that follow a pre-tool ack within the same
+    // turn, and consecutive answers in clips/talks flows. The orchestrator does not send a
+    // message id, so the previous id-mismatch heuristic does not apply.
+    const isNewAssistantMessage = lastChatEventType === 'answer' && lastMessage?.role === 'assistant';
```
dor-eitan: isnt the `lastChatEventType === 'answer'` enough?

src/services/socket-manager/message-queue.ts
```diff
@@ -58,7 +58,8 @@ function processChatEvent(
     chatEventQueue: ChatEventQueue,
     items: AgentManagerItems,
     onNewMessage: AgentManagerOptions['callbacks']['onNewMessage'],
-    clearQueue: () => void
+    clearQueue: () => void,
+    lastChatEventType: 'answer' | 'partial' | 'user'
```
dor-eitan: why not lastMessageType same as the caller?

src/services/socket-manager/message-queue.ts
```diff
@@ -144,7 +154,21 @@ export function createMessageEventQueue(
...
+                if (chatEvent === ChatProgress.Partial) {
+                    lastMessageType = 'partial';
+                } else if (chatEvent === ChatProgress.Answer) {
+                    lastMessageType = 'answer';
+                    // Clear the streaming buffer so the next turn's content does not inherit
+                    // leftover partial slots from this one.
+                    clearQueue();
+                } else if (chatEvent === ChatProgress.Transcribe) {
+                    lastMessageType = 'user';
+                }
```
dor-eitan: what about changing the lastMessageType to lastAssistantMessageType and change type to 'partial' | 'answer' | null

Then we will change this condition to else { lastAssistantMessageType = null } and we wont need to handle other events

src/services/socket-manager/message-queue.ts
dariusz-did: Good catch — the role check was redundant. A non-assistant `lastMessage` (user, transcribed user, or undefined) falls into the message-creation branch anyway, so dropping `&& lastMessage?.role === 'assistant'` is a pure simplification with no behavioural change. Done in 3adda87.

src/services/socket-manager/message-queue.ts
dariusz-did: Renamed both the parameter and the closure variable to `lastAssistantMessageType` in 3adda87.

src/services/socket-manager/message-queue.ts
dariusz-did: Applied in 3adda87 — variable renamed to `lastAssistantMessageType` with the narrower type `'partial' | 'answer' | null`. The explicit per-event handling (including `ChatProgress.Transcribe`) collapsed into a single `else { lastAssistantMessageType = null }` branch. The sentiment callback in the `StreamVideoCreated` handler now passes `lastAssistantMessageType ?? 'answer'` since its argument type is still `'answer' | 'partial' | 'user'`.

src/services/agent-manager/index.ts
dariusz-did: Updated approach in 80d961d — went with the simpler `if (response.result)` guard around the assistant push, as you suggested.

We tested follow-up turns live ("Can you tell me the weather..." + "nice, whats your name?") and the backend consistently returns `response.result === ""` for streaming modes — the streamed assistant reply arrives entirely through `chat/partial` + `chat/answer`. Skipping the push when `response.result` is empty removes the ghost assistant entries we were seeing in `items.messages` without needing the heavier `streamDeliveredThisTurn` check.

src/services/agent-manager/index.ts
dariusz-did: Resolved by going simpler in 80d961d. We no longer need to detect what the stream delivered — we just skip the push when `response.result` is empty, which handles the streaming case (no fallback content to write) without any extra bookkeeping. Textual modes still hit the push because the REST endpoint actually returns content there.

dariusz-did: You were right — the callback already had the `response.result` guard. With 80d961d both the push and the callback are gated by the same condition (`if (response.result)`), no redundant gates left.

dor-eitan (approved):

PR #399, opened 2026-05-31
fix(auth): rotate websocket connection id per session

osimhi213:
## Summary

- The trailing `_<id>` segment of the `Client-Key` auth header (formerly `sessionKey`) was generated **once at module load** and never rotated, so every session in the same tab reused the same value. PR #377 → #382 confirmed the backend uses this for per-tab WebSocket routing, so a stale-across-sessions value is wrong.
- Rename `sessionKey` → `websocketConnectionId`, expose `startSession()` from `src/auth/get-auth-header.ts`, and call it at the top of `agent-manager.connect()`. A new session gets a fresh id; retries within one `connect()` and any HTTP calls during that session share it (both go through `getAuthHeader`).
- No wire-format change. No new threading — the value lives in one module-scope variable with a single named rotation point.

## Test plan

- [x] `npx jest` — 414 tests pass, including the new `should rotate websocketConnectionId on startSession()` case.
- [x] `npx tsc --noEmit` — clean.
- [ ] Manual: open a session in the demo, verify the WS URL's `_<id>` suffix; reconnect (new session) and verify the suffix changes; check an HTTP request in the same session carries the same suffix.
- [ ] Backend log spot-check: confirm new sessions show up with distinct `connect_id` values.

### DIFF
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -14,6 +14,7 @@ import {
     SupportedStreamScript,
 } from '../../types';
 
+import { startSession } from '@sdk/auth/get-auth-header';
 import { CONNECTION_RETRY_TIMEOUT_MS } from '@sdk/config/consts';
 import { didApiUrl, didSocketApiUrl, mixpanelKey } from '@sdk/config/environment';
 import { ChatCreationFailed, ValidationError } from '@sdk/errors';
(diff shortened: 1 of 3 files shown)


src/services/agent-manager/index.ts
```diff
@@ -166,6 +167,7 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
     });
 
     async function connect(newChat: boolean) {
+        startSession();
```
dor-eitan: i would call it something like rotate connection id, its not starting anything

dor-eitan (approved):

src/services/agent-manager/index.ts
osimhi213: ok, thanks

PR #401, opened 2026-06-04
feat(analytics): track SDK errors to Mixpanel

osimhi213:
## What

- New `agent-error` Mixpanel event, tracked centrally by wrapping `options.callbacks.onError` in `createAgentManager` — every error surfaced by the SDK (api clients, socket manager, streaming managers, retry paths) is now tracked with its message. The consumer callback contract is unchanged.
- The existing `agent-message-send {event: 'error'}` now carries the error message too.
- `getErrorMessage(error)` util: error message as string, truncated to 256 chars (API error messages contain the raw response body).

## Why

Error events carried no detail, so production incidents couldn't be attributed from analytics. Investigating a real session today (LLM stalled 60s → chat POST timed out at the gateway's 29s cap → UI error handler switched to Maintenance and tore down a healthy WebRTC session) required a cross-system log dig; with this change the root cause is readable from the Mixpanel event (`error: '{"message":"Endpoint request timed out"}'`).

## Test plan

- `npx tsc --noEmit` clean
- Updated the chat-error assertion in `agent-manager/index.test.ts` (suite currently can't run due to a pre-existing `import.meta` jest config issue on main; `src/utils` suites pass)

### DIFF
--- a/src/utils/analytics.ts
+++ b/src/utils/analytics.ts
@@ -58,6 +58,8 @@ export function getAgentInfo(agent: Agent) {
         ...(agent.access === 'public' ? { from: 'agent-template' } : {}),
     };
 }
+export const getErrorMessage = (error: unknown) => String((error as Error)?.message ?? error ?? '').slice(0, 256);
+
 export const sumFunc = (numbers: number[]) => numbers.reduce((total, aNumber) => total + aNumber, 0);
 export const average = (numbers: number[]) => sumFunc(numbers) / numbers.length;
 
--- a/src/utils/analytics.ts
+++ b/src/utils/analytics.ts
@@ -58,7 +58,13 @@ export function getAgentInfo(agent: Agent) {
         ...(agent.access === 'public' ? { from: 'agent-template' } : {}),
     };
 }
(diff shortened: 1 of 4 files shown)


src/utils/analytics.ts
```diff
@@ -58,6 +58,8 @@ export function getAgentInfo(agent: Agent) {
         ...(agent.access === 'public' ? { from: 'agent-template' } : {}),
     };
 }
+export const getErrorMessage = (error: unknown) => String((error as Error)?.message ?? error ?? '').slice(0, 256);
```
dor-eitan: can it throw? should we wrap it with try/catch?

dor-eitan (approved):

src/utils/analytics.ts
osimhi213: Good catch — it could throw on values with no usable `toString` (e.g. `Object.create(null)`) or a throwing `.message` getter, and since it runs inside the error paths a throw there would mask the original error. Wrapped in try/catch with an `'Unknown error'` fallback and added a test for it in 4b6d2f0.

PR #408, opened 2026-06-09
feat(errors): typed, self-classifying errors (consumer contract + analytics)

osimhi213:
## Why

Errors reaching Mixpanel (`agent-error`, `agent-message-send`, `connectivity-error`) collapsed into a single opaque `"Failed to fetch"` string — a gateway timeout, a 402 insufficient-credits, an offline blip, a stream failure, and a 500 all looked identical. You couldn't tell what happened or which call failed. The same opacity hurt **consumers**: agents-ui recovered the kind via `JSON.parse(error.message)` / `instanceof TypeError` — fragile, message-coupled control flow.

## What

A small, self-classifying typed error model — both the **analytics** source *and* a **consumer-facing contract**. `message` is human-readable; machine logic branches on a stable `kind`.

- **`BaseError(message, kind?, originalError?)`** — concrete base; `kind` is the stable discriminant, `originalError` preserved as `cause`.
- **`HttpError(status, body, {url, method})`** — a non-2xx response; parses the server's `{ kind, description }` envelope, so it carries the server's own kind + `status`.
- **`NetworkError`** — transport failure: `fetch` rejected with no response (offline / DNS / refused / TLS / CORS). **Distinct from `HttpError`** — this is what finally makes "gateway timeout (504) vs Failed to fetch (no response)" distinguishable.
- **`StreamError`** — client-side stream/media failure (livekit / webrtc / socket). Previously plain `Error`s that analytics saw as `UnknownError`.
- **`ValidationError` / `WsError` / `ChatCreationFailed` / `ChatModeDowngraded`** — thin kinds on top of `BaseError`.
- **`isDIDError(e): e is BaseError`** — a field-based guard so consumers branch on `error.kind` robustly across bundlers / duplicated deps, where `instanceof` can silently fail (the `axios.isAxiosError` pattern).
- **`toErrorAnalytics(error)`** maps any thrown value to a lean, **allow-listed** payload: an SDK error's `toJson()` (`{ kind, message, cause?, httpStatus?, endpoint?, method? }`), or `{ kind: 'UnknownError', message }` for a non-SDK throwable.

**Invariant:** every error reaching analytics goes through `toErrorAnalytics`, nested under a single `error` property — `error.kind` for breakdowns, `error.message` for detail, namespaced away from event fields.

## Versioning: minor (`1.1.70` → `1.2.0`)

Additive to the *documented* public surface — new error classes + `isDIDError`. `error.message` was never a promised structure and "network rejects with `TypeError`" was incidental `fetch` behavior, so no documented contract is broken → **minor**, not major. (Registry is at `1.1.71`; `1.2.0` clears it.)

## Consumer note (behavioral change)

No *documented* contract changed, but code coupled to error internals needs a one-line migration. Both known consumers (agents-ui, agents-ui-2) are migrated + green (unit + a temporary integration test against the real SDK classes; type-checks clean against local `1.2.0`).

| Consumer read | Before | After |
|---|---|---|
| classify by | `JSON.parse(error.message).kind` | `isDIDError(e) ? e.kind : undefined` |
| network check | `e instanceof TypeError` | `isDIDError(e) && e.kind === 'NetworkError'` |
| `error.message` | raw `{kind,description}` JSON | the human description |
| `error.kind` / `error.status` | `undefined` | server kind / HTTP status |

Released in lockstep: SDK `1.2.0` + both UIs (they pin exact versions; the dep bump is the coordination point).

## Mixpanel payload — before → after

```jsonc
// before — one opaque string for everything
{ "error": "Failed to fetch" }

// after — HTTP error
{ "error": { "kind": "InsufficientCreditsError", "message": "no credits", "httpStatus": 402, "endpoint": "/agents/x/chat", "method": "POST" } }
// after — transport failure
{ "error": { "kind": "NetworkError", "message": "Network request failed", "cause": "Failed to fetch" } }
// after — stream/media failure
{ "error": { "kind": "StreamError", "message": "Track subscription timeout" } }
```

> ⚠️ The `error` property changes type **string → object** on these events. The old string values are the `"Failed to fetch"` noise this PR replaces — break down on **`error.kind`** going forward.

## Notes
- `apiClient` raises the typed errors; `createChat` propagates them untouched; the connect retry guard fail-fasts on `HttpError.status === 429` / `kind === 'InsufficientCreditsError'` (was message/status strings).
- The analytics payload is an explicit allow-list — `originalError`, request bodies, headers, and auth never leak.
- 456 tests (error classes, classifier, client incl. typed-contract, `createChat`, guard); `type-check` + prettier clean.

## Out of scope (pre-existing)
- The dead `retryHttpTooManyRequests` 429 retry (`fetch` resolves on 429, so it never fires).
- `NetworkError` doesn't carry `endpoint`/`method` yet.

### DIFF
--- a/src/services/chat/index.ts
+++ b/src/services/chat/index.ts
@@ -1,5 +1,4 @@
 import { PLAYGROUND_HEADER } from '@sdk/config/consts';
-import { HttpError } from '@sdk/errors';
 import type { Agent, AgentsAPI, Chat } from '@sdk/types';
 import { ChatMode } from '@sdk/types';
 import { isChatModeWithoutChat } from '@sdk/utils/chat';
@@ -17,28 +16,15 @@ export async function createChat(
     persist = false,
     chat?: Chat
 ) {
-    try {
-        if (!chat && !isChatModeWithoutChat(chatMode)) {
-            chat = await agentsApi.newChat(agent.id, { persist }, getRequestHeaders(chatMode));
+    if (!chat && !isChatModeWithoutChat(chatMode)) {
+        chat = await agentsApi.newChat(agent.id, { persist }, getRequestHeaders(chatMode));
 
-            analytics.track('agent-chat', {
-                event: 'created',
-                chatId: chat.id,
-                mode: chatMode,
-            });
-        }
-
-        return { chat, chatMode: chat?.chat_mode ?? chatMode };
-    } catch (error) {
-        const httpError = error instanceof HttpError ? error : undefined;
-
-        if (httpError?.kind === 'InsufficientCreditsError') {
-            throw new Error('InsufficientCreditsError');
-        }
-
-        // preserve the status so the connect-level retry guard can fail-fast on a 429
-        const failure = new Error('Cannot create new chat') as Error & { status?: number };
-        failure.status = httpError?.status;
-        throw failure;
+        analytics.track('agent-chat', {
+            event: 'created',
+            chatId: chat.id,
+            mode: chatMode,
+        });
     }
+
+    return { chat, chatMode: chat?.chat_mode ?? chatMode };
 }
--- a/src/services/chat/index.ts
+++ b/src/services/chat/index.ts
@@ -1,10 +1,11 @@
 import { PLAYGROUND_HEADER } from '@sdk/config/consts';
+import { HttpError } from '@sdk/errors';
 import type { Agent, AgentsAPI, Chat } from '@sdk/types';
 import { ChatMode } from '@sdk/types';
 import { isChatModeWithoutChat } from '@sdk/utils/chat';
 import { Analytics } from '../analytics/mixpanel';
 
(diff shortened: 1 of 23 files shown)


src/services/chat/index.ts
```diff
@@ -28,20 +29,16 @@ export async function createChat(
...
-const getErrorKind = (error: Error) => {
-    try {
-        const parsedError = JSON.parse(error.message);
-        return parsedError?.kind;
-    } catch (e) {
-        return 'UnknownError';
+        // preserve the status so the connect-level retry guard can fail-fast on a 429
+        const failure = new Error('Cannot create new chat') as Error & { status?: number };
+        failure.status = httpError?.status;
+        throw failure;
```
dor-eitan: what about thorwing a real error for 429?

dor-eitan (approved):

src/services/chat/index.ts
osimhi213: done.

dor-eitan (approved):

PR #410, opened 2026-06-12
feat(analytics): richer SDK error events (sdkVersion, emittedBy, NetworkError endpoint/method)

osimhi213:
## What

Makes SDK analytics events richer and easier to attribute in Mixpanel.

- **`emittedBy: 'agents-sdk'`** — static marker of the code that generated the event, distinct from `source` (a *host* heuristic: `agents-ui` vs standalone). When the SDK runs embedded, both SDK- and UI-emitted events report `source: 'agents-ui'`; `emittedBy` disambiguates them.
- **`sdkVersion`** — the SDK's published version, injected at build via a Vite `define` (`__SDK_VERSION__` ← `package.json`), surfaced as `SDK_VERSION` (`src/version.ts`) and re-exported from the package entry. Falls back to `'dev'` when the define isn't applied (jest/dev).
- **`NetworkError` now carries `endpoint` + `method`** — a `NetworkError` (fetch rejected with no response) previously serialized only `{kind, message, cause}`, so you couldn't tell *which* request failed on prod. The url/method are already known at the throw site (`apiClient` passes them to `onError`); now they're baked into the error too, so `toJson()` emits `endpoint`/`method` — matching `HttpError`'s shape.

## Why

Error events couldn't be attributed to the emitting codebase, carried no SDK version, and — for the most common prod failure (`NetworkError`) — didn't say which endpoint failed.

## The publish-workflow fix (important)

`publish-on-merge.yml` built the bundle **before** determining/writing the new version, so the baked `SDK_VERSION` would have been the *pre-bump* value (`1.2.0` while publishing `1.2.0-staging.N` / `1.2.1`). Reordered to **Determine version → Update package.json → Build → Publish**.

## Resulting events

| context | `source` (unchanged) | `emittedBy` (new) |
|---|---|---|
| SDK, embedded in UI | `agents-ui` | `agents-sdk` |
| SDK, standalone | `agents-sdk` | `agents-sdk` |

A `NetworkError` now serializes e.g.:
```json
{ "kind": "NetworkError", "message": "Network request failed", "cause": "Failed to fetch", "endpoint": "/agents/x/chat", "method": "POST" }
```

## Test plan

- `yarn type-check` clean; full suite **459/459** passing (added `NetworkError` toJson coverage + an end-to-end apiClient assertion that endpoint/method propagate).
- Built locally; confirmed `__SDK_VERSION__` is fully replaced and `SDK_VERSION` bakes to the `package.json` version.

### DIFF
--- a/src/services/analytics/mixpanel.ts
+++ b/src/services/analytics/mixpanel.ts
@@ -1,5 +1,6 @@
 import { getExternalId } from '@sdk/auth/get-auth-header';
 import { getRandom } from '@sdk/utils';
+import { SDK_VERSION } from '@sdk/version';
 
 export interface AnalyticsOptions {
     token: string;
@@ -77,6 +78,8 @@ export function initializeAnalytics(config: AnalyticsOptions): Analytics {
                                 ...sendProps,
                                 agentId: this.agentId,
                                 source,
+                                emittedBy: 'agents-sdk',
+                                sdkVersion: SDK_VERSION,
                                 token: this.token,
                                 time: eventTime,
                                 $insert_id: this.getRandom(),
--- a/.github/workflows/publish-on-merge.yml
+++ b/.github/workflows/publish-on-merge.yml
@@ -75,12 +75,6 @@ jobs:
           yarn install --frozen-lockfile \
             || logger -l error -m "Failed to install dependencies"
 
-      - name: Build package
-        run: |
-          logger -l info -m "Building package"
-          yarn build \
-            || logger -l error -m "Failed to build package"
-
       - name: Determine version and tag
         id: version
         run: |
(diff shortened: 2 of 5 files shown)


src/services/analytics/mixpanel.ts
```diff
@@ -77,6 +78,8 @@ export function initializeAnalytics(config: AnalyticsOptions): Analytics {
                                 ...sendProps,
                                 agentId: this.agentId,
                                 source,
+                                emittedBy: 'agents-sdk',
```
dor-eitan, line 81: whats the diff between this and source?

dor-eitan (approved):

src/services/analytics/mixpanel.ts
osimhi213: source is which user are you: user using agents-sdk (developer) or user that uses agents-ui (probably embed/studio)
emittedBy is basically from where this event is coming from. bascially the 'source' lol.

PR #417, opened 2026-06-16
feat(stats): A/V-sync observability + anomalies in the WebRTC stats report

osimhi213:
## What
Adds audio/video lip-sync observability to the WebRTC stats pipeline, folded into the existing `VideoRTCStatsReport` so it rides the current `stream-session`/`agent-video` analytics — **no new event, no second `getStats()` loop**.

## Why
The SDK already tracks per-session video stats (rtt, jitter, freezes, packet loss) but has **no audio or A/V-sync visibility** — it can't see lip-sync drift, the user-facing "audio ahead of video at session start" problem. This surfaces it as structured metrics + named anomalies.

## How
- **`formatStats`** now also reads the audio `inbound-rtp` and `remote-outbound-rtp` (RTCP Sender Report) entries per tick, stored on `SlimRTCStatsReport.av` — reusing the existing 100 ms video-stats loop.
- **`stats/av-sync.ts`** (new):
  - `buildAvSyncReport(stats)` derives the per-session summary: rendered A/V offset (median / p95 / steady-state residual), time-to-sync, start offset, sync buffering (audio JB target−min), audio↔video JB gap, audio retiming, SR clock ratio + skew, conceal, audio loss/jitter.
  - `detectAvSyncAnomalies()` classifies named, severity-tagged anomalies: `broken-sr-clock`, `unbounded-drift`, `never-converged`, `slow-start-convergence`, `residual-drift`, `sr-clock-skew`, `excess-sync-buffering`.
- **`createVideoStatsReport`** attaches `webRTCStats.avSync: AvSyncReport | null`.

Because `connect-to-manager` already spreads `...statsReport` into `stream-session`/`agent-video`, `avSync` (incl. its `anomalies`) reaches Mixpanel automatically.

## Behaviour / graceful degradation
- `avSync` is `null` when there aren't ≥2 paired audio+video playout samples (audio-only / pre-roll).
- SR-based fields (`srClockRatio*`, `srSkewMs`) are `null` on browsers that don't expose `remote-outbound-rtp` (non-Chromium); `broken-sr-clock` is skipped in that case (no false positives).

## Validation
Dev-vs-prod A/B against the media-gateway A/V-sync fix:
- **prod (unfixed):** A/V offset 19k–48k ms, never converges → trips `unbounded-drift` + `never-converged` (+ `broken-sr-clock` when SRs present).
- **dev (fixed):** converges to ~95 ms in ~3.5 s → at most `slow-start-convergence`.

### DIFF
--- a/src/services/streaming-manager/stats/av-sync.ts
+++ b/src/services/streaming-manager/stats/av-sync.ts
@@ -1,0 +1,135 @@
+import { AvSyncAnomaly, AvSyncReport, AvSyncSample, SlimRTCStatsReport } from '@sdk/types';
+
+const CONVERGED_THRESHOLD_MS = 100;
+
+const round = (value: number) => Math.round(value);
+const round2 = (value: number) => Math.round(value * 100) / 100;
+const percentile = (sorted: number[], p: number) =>
+    sorted[Math.min(sorted.length - 1, Math.floor(p * (sorted.length - 1)))];
+
+function detectAvSyncAnomalies(report: Omit<AvSyncReport, 'anomalies'>): AvSyncAnomaly[] {
+    const anomalies: AvSyncAnomaly[] = [];
+    const add = (condition: boolean, type: string, severity: AvSyncAnomaly['severity'], detail: string) => {
+        if (condition) {
+            anomalies.push({ type, severity, detail });
+        }
+    };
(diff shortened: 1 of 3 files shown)


osimhi213:
Closing — rolling back to review locally before pushing.

PR #438, opened 2026-07-14
feat: optional sentiment param in speak

michaelyanetzkoren:
### Pull Request Type

🔮 Feature

### General Description

- `speak` now accepts an optional `sentiment` name on text scripts (expressive avatars)
- Unsupported sentiments fall back to the agent's default sentiment on the server side

### Technical Description

- Added optional `sentiment` field to `Stream_Text_Script`
- `speak()` preserves `sentiment` when it fills in the default voice provider; scripts with an explicit provider already pass through unchanged
- README documents the param with an example; two unit tests cover sentiment passthrough

### DIFF
--- a/src/types/stream-script.ts
+++ b/src/types/stream-script.ts
@@ -33,7 +33,7 @@ export interface Stream_Text_Script extends BaseStreamScript {
     ssml?: boolean;
 
     /**
-     * Queue this speak behind the current speech instead of interrupting it.
+     * Queue this speak behind the current speech instead of interrupting it (expressive avatars only).
      * @default false
      */
     should_queue_speaks?: boolean;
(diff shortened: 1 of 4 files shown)


src/types/stream-script.ts
```diff
@@ -37,6 +37,13 @@ export interface Stream_Text_Script extends BaseStreamScript {
      * @default false
      */
     should_queue_speaks?: boolean;
+
+    /**
+     * Sentiment name to speak with (expressive avatars only).
+     * If the sentiment is not supported by the agent, the default sentiment is used.
+     * @example "friendly"
+     */
+    sentiment?: string;
```
osimhi213, line 46: senitment and should_queue_speaks should exist only on expressive scripts

osimhi213 (approved):

PR #442, opened 2026-07-21
feat!: minimal RuntimeAgent for the embedded widget (breaking, 2.0.0)

dor-eitan:
### Pull Request Type

🔮 Feature

### Description

-   **BREAKING (2.0.0):** `createAgentManager().agent` (`manager.agent`) is now a minimal **`RuntimeAgent`** instead of the full `Agent`. The widget loads via the new `GET /agents/{id}/runtime` endpoint, which returns only the fields the widget needs and drops internals (`llm` / system prompt, `tools`, raw `triggers`, etc.).
-   Adds `agentsApi.getRuntimeById()` → `RuntimeAgent`. `agentsApi.getById()` and the full `Agent` type are **unchanged** for consumers that still need the full agent.
-   `speak()` no longer sends the TTS voice (`script.provider`) — the backend/worker resolves it from the stored agent — and no longer throws `"Presenter voice is not initialized"`. An explicitly-supplied `provider` is still honored.
-   `starter_message` is now read top-level (was `knowledge.starter_message`). V1/V2 stream routing (`isStreamsV2Agent`) is unchanged.
-   Analytics helpers trimmed to the `RuntimeAgent` shape (agent-config props like `llm.*` / `voice_id` dropped), now with direct unit tests. `yarn test` 498 green; `yarn type-check` clean.
-   Depends on the backend runtime endpoint — serverless PR #7377.

### Reference Links

-   [Asana](https://app.asana.com/1/856614567666442/project/1216229456433893/task/1216381104254852?focus=true)


src/services/agent-manager/connect-to-manager.test.ts
```diff
@@ -679,7 +675,7 @@ describe('connect-to-manager', () => {
                     presenter_id: 'expressive-presenter-123',
                     voice: { type: Providers.Microsoft, voice_id: 'voice-123' },
                 },
-            };
+            } as Agent;
```
dor-eitan: why do we need that casting?

src/test-utils/factories/agents-api.factory.ts
```diff
@@ -4,6 +4,12 @@ import { AgentFactory } from './agent.factory';
 
 export const AgentsApiFactory = new Factory().attrs({
     getById: () => jest.fn().mockResolvedValue(AgentFactory.build()),
+    getRuntimeById: () => {
+        // The /agents/{id}/runtime payload (RuntimeAgent) carries starter_message at the top level,
+        // not nested under knowledge as the base Agent does.
```
dor-eitan: dont need these comments

src/services/agent-manager/connect-to-manager.test.ts
dor-eitan: Retyped the fixtures from `as Agent` to `as RuntimeAgent` — they were still typed to the old `Agent` since `manager.agent` is now `RuntimeAgent`. The cast is retained (not dropped) because these shared fixtures intentionally carry a few props beyond `RuntimeAgent` (e.g. `presenter_id`, `knowledge.provider`) that other cases in this file exercise. If you'd rather have cast-free fixtures, I can trim them to the exact `RuntimeAgent` shape — let me know.

src/test-utils/factories/agents-api.factory.ts
dor-eitan: Removed the comments.

src/services/agent-manager/connect-to-manager.test.ts
dor-eitan: Good call — removed the cast entirely. Trimmed both fixtures to the exact `RuntimeAgent` shape (the leftover `Agent`-era props — `name`, `presenter.driver_id`/`presenter_id`/`is_greenscreen`/`background`, `knowledge.provider`, `voice_id` — were never read by the code or asserted by any test), so `let mockAgent: RuntimeAgent = { ... }` type-checks with no `as`. Dropped the now-unused `Providers` import too. type-check clean, 500/500 tests green.

src/types/entities/agents/agent.ts
```diff
@@ -80,6 +81,7 @@ export interface AgentsAPI {
     create(payload: AgentPayload, options?: RequestInit): Promise<Agent>;
     getAgents(tag?: string, options?: RequestInit): Promise<Agent[]>;
     getById(id: string, options?: RequestInit): Promise<Agent>;
```
ariksfaradi: So who uses it?

src/types/entities/agents/runtime-agent.ts
```diff
@@ -0,0 +1,18 @@
+import { EndOfCallFeedbackConfig } from './agent';
+
+export interface RuntimeAgent {
```
ariksfaradi: maybe just change the interface of RuntimeAgent, drop the getById (leave only getRuntimeAgent -> return Agent with the new struct)
and because its a major upgrade add "migrate from v1 to v2" md file

ariksfaradi (approved):

src/types/entities/agents/agent.ts
dor-eitan: Nothing did — `getById` (the full-agent fetch) had no internal callers and wasn't exposed to consumers (the `agentsApi` instance is never returned; only the type was exported). Removed it, along with the equally-dead `create`/`update`/`getAgents` + `AgentPayload`. `getRuntimeById` is now the sole agent fetch, returning the new `Agent`. (954ec6cf)

src/types/entities/agents/runtime-agent.ts
dor-eitan: Done, and went with your approach: folded `RuntimeAgent` into `Agent` — in the SDK's scope, `Agent` *is* the runtime agent (the minimized shape). Dropped the dead `getById`/`create`/`update`/`getAgents` + `AgentPayload` + the old full `Agent`; kept `getRuntimeById`. Added `MIGRATION.md` (v1→v2 field-rename table + removed-API list + the `speak()` no-`provider` note). type-check clean, 500/500 tests. (954ec6cf)

PR #444, opened 2026-07-26
feat(turn-events): consume turn/started + turn/ended for agent activity state

osimhi213:
### Pull Request Type

🔮 Feature

### Description
- Consume the new `turn/started` / `turn/ended` events and drive `AgentActivityState` from the **turn lifecycle** instead of inferring it per sequence: `turn/started` → `Loading`, `turn/ended` → the **sole** `Idle` terminal.
- A `usingTurnEvents` latch keeps servers that don't emit turn events on the exact legacy per-sequence behavior (fully backward compatible).
- Mid-turn `stream-video/*` and `tool-call/*` terminals now **stay busy** (`Talking` / `ToolActive`) instead of flicking to `Idle` — fixes the mid-turn idle flicker on multi-sequence (tool) turns.
- A stale `turn/ended` from a superseded turn is dropped (`turn_id < currentTurnId`); `turn_id` is mirrored from the server, never counted client-side, so it can't drift.
- Adds `StreamEvents.TurnStarted` / `TurnEnded` and 4 unit tests (Loading on start, Idle only on end, stay-ToolActive mid-turn, straggler drop).

Server contract: de-id/streaming-orchestrator-worker#583.

### Reference Links
-   [Asana]()

### DIFF
--- a/src/services/streaming-manager/livekit-manager.ts
+++ b/src/services/streaming-manager/livekit-manager.ts
@@ -126,6 +126,8 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
     let currentActivityState: AgentActivityState = AgentActivityState.Idle;
     let currentInterruptible = true;
     const pendingToolCalls = new Set<string>();
+    let currentTurnId: number | null = null;
+    let usingTurnEvents = false;
 
     const streamApi = createStreamApiV2(auth, baseURL || didApiUrl, agentId, callbacks.onError);
     let sessionId: string | undefined;
@@ -172,7 +174,7 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
     function handleTranscriptionReceived(_segments: TranscriptionSegment[], participant?: Participant): void {
         if (participant?.isLocal) {
             latencyTimestampTracker.update();
-            if (currentActivityState === AgentActivityState.Talking) {
+            if (!usingTurnEvents && currentActivityState === AgentActivityState.Talking) {
                 currentActivityState = AgentActivityState.Idle;
             }
         }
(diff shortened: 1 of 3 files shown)


src/services/streaming-manager/livekit-manager.ts
```diff
@@ -469,6 +471,20 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
         });
     }
 
+    function handleTurnStarted(_subject: string, data: any): void {
```
dor-eitan: replace any

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -469,6 +471,20 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
...
+        usingTurnEvents = true;
+        currentTurnId = data?.turn_id ?? null;
+        currentActivityState = AgentActivityState.Loading;
+        callbacks.onAgentActivityStateChange?.(AgentActivityState.Loading);
+    }
+
+    function handleTurnEnded(_subject: string, data: any): void {
```
dor-eitan: replace any

dor-eitan (approved):

src/services/streaming-manager/livekit-manager.ts
osimhi213: > replace any

Typed both turn handlers with `TurnEventPayload { turn_id: number | null }` (added to stream.ts) in 8afcd86.

_authored by claude using pr:review@de-id skill_

src/services/streaming-manager/livekit-manager.ts
osimhi213: > replace any

Typed both turn handlers with `TurnEventPayload { turn_id: number | null }` (added to stream.ts) in 8afcd86.

_authored by claude using pr:review@de-id skill_

PR #457, opened 2026-08-09
fix(tools): derive interruptibility from all pending calls

dor-eitan:
### Pull Request Type

🐛 BugFix

### Description

-   Derives interruptibility from all pending tool calls instead of the last event — resolves the TODO at `livekit-manager.ts`.
-   Exposes the running calls: `onRunningToolCallsChange(calls: readonly RunningToolCall[])` plus an exported `isAwaitingTool(calls)`. `RunningToolCall` is `{ callId, name, executionMode }`.

Neither `ToolActive` nor `isInterruptible` can distinguish "an async tool is open" from "the tool finished and the answer is coming" — `ToolActive` stays true until the turn ends, and an empty set reads as interruptible. That gap dropped the thinking indicator after every tool call (see de-id/agents-ui#1092).

### Technical notes

-   The aggregate is sourced from `executionMode`, not `interruptible` — which returns to meaning cancellable in de-id/streaming-orchestrator-worker#635. `onInterruptibleChange` keeps its public meaning.
-   Speech interruptibility dropped as an input: it duplicated the pending-call map with different timing.
-   Stale blocking calls are dropped at `turn/ended`, so a missing terminal event can't pin interruptibility for the session.
-   Per-call `interruptible` stays internal — it's the future cancellation flag, and no consumer needs it yet.

### Tests

`yarn test` → 526/526; `tsc --noEmit` and prettier clean.

### Rollout

Release before de-id/agents-ui#1092 — it imports `isAwaitingTool` and `RunningToolCall`, so its build fails until this ships and the pin is bumped.

### Reference Links

-   [Asana](https://app.asana.com/1/856614567666442/project/1216554029921535/task/1216473369651534?focus=true)

### DIFF
--- a/src/services/streaming-manager/livekit-manager.ts
+++ b/src/services/streaming-manager/livekit-manager.ts
@@ -53,6 +53,12 @@ interface TrackPublishState {
     publication: LocalTrackPublication | null;
 }
 
+interface PendingToolCall {
+    call: RunningToolCall;
+    interruptible: boolean;
+    turnId: number | null;
+}
+
 async function importLiveKit(): Promise<{
     Room: typeof Room;
     RoomEvent: typeof RoomEvent;
@@ -130,7 +136,7 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
     let trackSubscriptionTimeoutId: ReturnType<typeof setTimeout> | null = null;
     let currentActivityState: AgentActivityState = AgentActivityState.Idle;
     let currentInterruptible = true;
-    const pendingToolCalls = new Map<string, RunningToolCall & { interruptible: boolean; turnId: number | null }>();
+    const pendingToolCalls = new Map<string, PendingToolCall>();
     let currentTurnId: number | null = null;
 
     const streamApi = createStreamApiV2(auth, baseURL || didApiUrl, agentId, callbacks.onError);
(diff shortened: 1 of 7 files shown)


src/services/streaming-manager/livekit-manager.ts
```diff
@@ -126,7 +129,7 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
     let trackSubscriptionTimeoutId: ReturnType<typeof setTimeout> | null = null;
     let currentActivityState: AgentActivityState = AgentActivityState.Idle;
     let currentInterruptible = true;
-    const pendingToolCalls = new Set<string>();
+    const pendingToolCalls = new Map<string, RunningToolCall & { interruptible: boolean; turnId: number | null }>();
```
dor-eitan: change to proper interface

dor-eitan: Done in `4694532`.

Went with the nested shape rather than a flat rename:

```ts
interface PendingToolCall {
    call: RunningToolCall;
    interruptible: boolean;
    turnId: number | null;
}
```

Building `call` once at `tool-call/started` and emitting `entries.map(e => e.call)` makes "the private fields never leak into the emitted collection" a type-level guarantee instead of a hand-maintained strip list — otherwise adding a field to `RunningToolCall` would silently stop it being emitted. Kept unexported; it is internal state, not public API.

dariusz-did (approved):

PR #462, opened 2026-08-20
feat: presentation data channel topic and generic sendDataMessage

dor-eitan:
### Pull Request Type

🔮 Feature

### Description

- Presenter Mode client machinery: `DataChannelTopic.Presentation = 'did.presentation'` and a generic `sendDataChannelMessage(topic, payload)` on the `AgentManager` (JSON over the LiveKit text-stream path, same as `did.stt-language`), consumed by agents-ui's presenter UI via dev:local-sdk. Documented in the README alongside the other Expressive-only methods.
- Send-surface cleanup (review-driven): the `StreamingManager` type now has ONE send primitive, `sendDataChannelMessage(topic: DataChannelTopic, payload: string): Promise<void>`, implemented by both transports — LiveKit routes it over the topic'd text stream; WebRTC sends the payload as-is and ignores the topic (it has no topic concept). LiveKit's private `sendMessage` twin is gone (the public method is the implementation), and the members that were only a topic in disguise moved to their call sites: the V2 chat path sends on `lk.chat` and the agent-manager's `setSttLanguage` sends on `did.stt-language`, so `sendTextMessage` and `StreamingManager.setSttLanguage` are deleted. Only `speak` and `interrupt` keep members, because their mechanics genuinely differ per transport (V1 `speak` posts a stream request; V1 `interrupt` sends a `stream/interrupt` payload where V2 sends an empty `did.interrupt`), and everything they put on the data channel goes through the primitive. The old LiveKit `sendDataChannelMessage(payload)` wrapper is deleted — it JSON-parsed the payload only to read `.topic` and then sent an empty body (a latent bug), and the test-only intersection type on the WebRTC factory is gone.
- `DataChannelTopic` moved out of `livekit-manager.ts` into `src/types/stream` next to `StreamEvents`, which removes the types→service import edge and the cycle that typing the primitive would otherwise have created; the package root still exports it, so the public surface is unchanged. The topic is now typed `DataChannelTopic` (not `string`) on the shared type, both implementations and the AgentManager method.
- Type-surface changes for SDK consumers: `StreamingManager`'s send-ish surface is now `sendDataChannelMessage` + `speak` + `interrupt`; `DataChannelTopic` is exported from the package root; `AgentManager` gains a required `sendDataChannelMessage(topic, payload)` — same name as the transport primitive it stringifies into, one concept with one name at both layers. `AgentManager.setSttLanguage` keeps its signature and its promise contract: it resolves when the send settles, as it did on main. Wire behavior is byte-identical on both transports (pinned by the existing sendText / dataChannel.send assertions).
- Regression caught in review and fixed (`cc56144`): moving the call sites had dropped two `await`s against LiveKit's async send — the V2 chat path and `setSttLanguage`, which had started returning `Promise.resolve()` — because the shared type declared the primitive `void`; it is now typed `Promise<void>`, WebRTC's implementation is async to match, and both callers await/return the send's promise again (the promise tracks the send attempt and never rejects — failures route to `onError`, on main too). `sendDataChannelMessage` is gated to Expressive (V4) agents as its doc says (`1d96caa`).
- Analytics: `agent-data-message` Mixpanel event with `{ topic }` (mirroring `setSttLanguage`); the event name stays as-is for analytics consumers even after the method rename. The duplicate UI-level navigate event was removed in agents-ui.
- Tests: `yarn ci:test` 536/536 (534 + two pins that the send is awaited); coverage above all thresholds; lint clean.
- Release path: merge to main publishes a `staging`-tagged version; agents-ui #1105 no-ops safely on the published 2.0.6 pin and bumps after promotion.

### Reference Links

-   [Asana](https://app.asana.com/1/856614567666442/project/1216943078884032/task/1216943731956184?focus=true)

https://claude.ai/code/session_01CJftELCLQpj4MUq1Z86JwB

### DIFF
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -34,7 +34,7 @@ import { getInitialMessages } from '../chat/intial-messages';
 import { SocketManager, createSocketManager } from '../socket-manager';
 import { createMessageEventQueue } from '../socket-manager/message-queue';
 import { StreamingManager } from '../streaming-manager';
-import type { DataChannelTopic } from '../streaming-manager/livekit-manager';
+import { DataChannelTopic } from '../streaming-manager/livekit-manager';
 import { initializeStreamAndChat } from './connect-to-manager';
 
 export interface AgentManagerItems {
@@ -344,13 +344,15 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
             return items.streamingManager.publishMicrophoneStream(stream);
         },
         setSttLanguage(language: string): Promise<void> {
-            if (!items.streamingManager?.setSttLanguage) {
+            if (!isStreamsV2 || !items.streamingManager) {
                 return Promise.reject(new Error('setSttLanguage is not available for this streaming manager'));
             }
 
             analytics.track('agent-stt-language-change', { language });
 
-            return items.streamingManager.setSttLanguage(language);
+            items.streamingManager.sendDataChannelMessage(DataChannelTopic.SttLanguage, JSON.stringify({ language }));
+
+            return Promise.resolve();
         },
         sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
             if (!items.streamingManager?.sendDataChannelMessage) {
@@ -435,7 +437,7 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
 
                 const chatRequestFn = useV2Path
                     ? async () => {
-                          await items.streamingManager?.sendTextMessage?.(userMessage);
+                          items.streamingManager?.sendDataChannelMessage(DataChannelTopic.Chat, userMessage);
                           return Promise.resolve({} as ChatResponse);
                       }
                     : async () => {
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -353,13 +353,15 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
             return items.streamingManager.setSttLanguage(language);
         },
         sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
-            if (!items.streamingManager?.sendDataMessage) {
-                return Promise.reject(new Error('sendDataMessage is not available for this streaming manager'));
+            if (!items.streamingManager?.sendDataChannelMessage) {
+                return Promise.reject(new Error('sendDataChannelMessage is not available for this streaming manager'));
             }
 
             analytics.track('agent-data-message', { topic });
 
-            return items.streamingManager.sendDataMessage(topic, payload);
+            items.streamingManager.sendDataChannelMessage(topic, JSON.stringify(payload));
+
+            return Promise.resolve();
         },
         unpublishMicrophoneStream(): Promise<void> {
             if (!items.streamingManager?.unpublishMicrophoneStream) {
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -357,6 +357,8 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
                 return Promise.reject(new Error('sendDataMessage is not available for this streaming manager'));
             }
 
+            analytics.track('agent-data-message', { topic });
+
             return items.streamingManager.sendDataMessage(topic, payload);
         },
         unpublishMicrophoneStream(): Promise<void> {
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -355,8 +355,8 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
             return Promise.resolve();
         },
         sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
-            if (!items.streamingManager?.sendDataChannelMessage) {
-                return Promise.reject(new Error('sendDataChannelMessage is not available for this streaming manager'));
+            if (!isStreamsV2 || !items.streamingManager) {
+                return Promise.reject(new Error('sendDataMessage is not available for this streaming manager'));
             }
 
             analytics.track('agent-data-message', { topic });
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -1,3 +1,4 @@
+import { DataChannelTopic } from '@sdk/types/stream/data-channel';
 import {
     AgentManager,
     AgentManagerOptions,
@@ -7,9 +8,9 @@ import {
     ClientToolHandler,
     ConnectionState,
     CreateStreamOptions,
-    DataChannelTopic,
     Interrupt,
     Message,
+    PublicDataChannelTopic,
     StreamScript,
     SupportedStreamScript,
 } from '../../types';
@@ -355,7 +356,7 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
                 JSON.stringify({ language })
             );
         },
-        sendDataChannelMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
+        sendDataChannelMessage(topic: PublicDataChannelTopic, payload: Record<string, unknown>): Promise<void> {
             if (!isStreamsV2 || !items.streamingManager) {
                 return Promise.reject(new Error('sendDataChannelMessage is not available for this streaming manager'));
             }
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -350,9 +350,10 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
 
             analytics.track('agent-stt-language-change', { language });
 
-            items.streamingManager.sendDataChannelMessage(DataChannelTopic.SttLanguage, JSON.stringify({ language }));
-
-            return Promise.resolve();
+            return items.streamingManager.sendDataChannelMessage(
+                DataChannelTopic.SttLanguage,
+                JSON.stringify({ language })
+            );
         },
         sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
             if (!isStreamsV2 || !items.streamingManager) {
@@ -361,9 +362,7 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
 
             analytics.track('agent-data-message', { topic });
 
-            items.streamingManager.sendDataChannelMessage(topic, JSON.stringify(payload));
-
-            return Promise.resolve();
+            return items.streamingManager.sendDataChannelMessage(topic, JSON.stringify(payload));
         },
         unpublishMicrophoneStream(): Promise<void> {
             if (!items.streamingManager?.unpublishMicrophoneStream) {
@@ -437,7 +436,7 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
 
                 const chatRequestFn = useV2Path
                     ? async () => {
-                          items.streamingManager?.sendDataChannelMessage(DataChannelTopic.Chat, userMessage);
+                          await items.streamingManager?.sendDataChannelMessage(DataChannelTopic.Chat, userMessage);
                           return Promise.resolve({} as ChatResponse);
                       }
                     : async () => {
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -355,9 +355,9 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
                 JSON.stringify({ language })
             );
         },
-        sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
+        sendDataChannelMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
             if (!isStreamsV2 || !items.streamingManager) {
-                return Promise.reject(new Error('sendDataMessage is not available for this streaming manager'));
+                return Promise.reject(new Error('sendDataChannelMessage is not available for this streaming manager'));
             }
 
             analytics.track('agent-data-message', { topic });
--- a/src/services/agent-manager/index.ts
+++ b/src/services/agent-manager/index.ts
@@ -34,6 +34,7 @@ import { getInitialMessages } from '../chat/intial-messages';
 import { SocketManager, createSocketManager } from '../socket-manager';
 import { createMessageEventQueue } from '../socket-manager/message-queue';
 import { StreamingManager } from '../streaming-manager';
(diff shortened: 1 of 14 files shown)


src/services/streaming-manager/livekit-manager.ts
```diff
@@ -873,6 +874,16 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
...
 
+        /**
+         * Send a JSON payload to the agent over a data-channel topic.
+         * @param topic - The topic to send on.
+         * @param payload - Plain object, serialized as JSON.
+         * @returns Promise that resolves after the send attempt completes; failures are reported via onError.
+         */
+        sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>) {
+            return sendMessage(JSON.stringify(payload), topic);
+        },
```
dor-eitan, line 842: we have now sendDataChannelMessage, sendDataMessage, sendMessage, sendTextMessage - why do we need so much? lets clean it

src/services/agent-manager/index.ts
```diff
@@ -351,6 +352,13 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
 
             return items.streamingManager.setSttLanguage(language);
         },
+        sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
+            if (!items.streamingManager?.sendDataMessage) {
+                return Promise.reject(new Error('sendDataMessage is not available for this streaming manager'));
+            }
+
+            return items.streamingManager.sendDataMessage(topic, payload);
```
dor-eitan: should we send a mixpanel event like other similar methods here?

src/services/streaming-manager/livekit-manager.ts
dor-eitan: Cleaned up in `bcc1526`: `sendDataChannelMessage` is gone from the LiveKit manager — it parsed the payload only to read `.topic` and then sent an empty body, so it was a latent bug on top of being a duplicate. `interrupt()` now goes through the internal `sendMessage('', Interrupt)`, and the member is optional on the type: WebRTC still needs it for raw `RTCDataChannel` strings, kept precise via an intersection type on the V1 factory.

`sendTextMessage` stays — it's the raw-string sender, distinct from the JSON `sendDataMessage`.

src/services/agent-manager/index.ts
dor-eitan: Added in `cad1f07` — `agent-data-message` with `{ topic }`, mirroring `setSttLanguage`.

Heads-up as discussed: agents-ui also tracks its own `agent-presentation-user-navigate`, so a navigate will show up twice in Mixpanel (SDK-level + UI-level).

dor-eitan: so lets remove it from the ui level

src/services/streaming-manager/common.ts
```diff
@@ -112,6 +114,14 @@ export type StreamingManager<T extends CreateStreamOptions | CreateSessionV2Opti
      */
     setSttLanguage?(language: string): Promise<void>;
 
+    /**
+     * Send a JSON payload to the agent over a data-channel topic.
+     * supported only for livekit manager
+     * @param topic Data-channel topic to send on
+     * @param payload Plain object, serialized as JSON
+     */
+    sendDataMessage?(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void>;
```
dor-eitan: why its not sendDataChannelMessage? why one for livekit and one for webrtc

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -873,6 +862,16 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
             return sendMessage(JSON.stringify({ language }), DataChannelTopic.SttLanguage);
```
dor-eitan: this and others should use the sendDataChannelMessage

src/services/streaming-manager/webrtc-manager.ts
```diff
@@ -149,7 +149,7 @@ export async function createWebRTCStreamingManager<T extends CreateStreamOptions
     streamOptions: T,
     { debug = false, callbacks, auth, baseURL = didApiUrl, analytics }: StreamingManagerOptions,
     signal?: AbortSignal
-): Promise<StreamingManager<T>> {
+): Promise<StreamingManager<T> & { sendDataChannelMessage(payload: string): void }> {
```
dor-eitan: not good

src/services/agent-manager/index.ts
dor-eitan: Removed in agents-ui 9517b65b (PR de-id/agents-ui#1105) — the `trackUi('agent-presentation-user-navigate')` call is gone; the navigate send itself stays.

One consequence to be aware of: the SDK event carries only `{topic}`, so slide-number granularity leaves Mixpanel (the UI event carried `{slide}`). Topic-level navigate counts remain, and re-adding slide granularity later is a one-liner if analytics needs it.

src/services/streaming-manager/webrtc-manager.ts
dor-eitan: Agreed, removed in 39725d7. The intersection was production-dead anyway: `factory.ts` declares the return as the plain `StreamingManager<...>`, so it got erased for every real caller, and it contradicted the `WebRTCStreamingManager<T> = StreamingManager<T>` alias right below. `sendDataChannelMessage` is now a required member on the shared type, so nothing needs an intersection - the factory returns `Promise<StreamingManager<T>>` and the direct-import V1 tests type-check plainly.

src/services/streaming-manager/common.ts
dor-eitan: Unified in 39725d7. There is now one required method on the shared type, replacing both optionals: `sendDataChannelMessage(topic: string, payload: string): void`. The transports differ only in how they use it, and the type docs say so: LiveKit sends it as a topic'd text stream (`sendText(payload, { topic })`), WebRTC sends the payload as-is on the RTC data channel and ignores the topic, having no topic concept. Same wire bytes as before on both sides. `AgentManager.sendDataMessage(topic, payload)` stays as the app-facing API and now delegates via `sendDataChannelMessage(topic, JSON.stringify(payload))`.

src/services/streaming-manager/livekit-manager.ts
dor-eitan: Done in 39725d7. Note the literal `sendDataChannelMessage` this line referred to had been deleted earlier in the PR at your request (bcc1526, "drop the livekit sendDataChannelMessage wrapper") - this commit restores the name as THE public method on the shared type, and every send funnels through it: `interrupt` (`did.interrupt`, empty payload), `setSttLanguage` (`did.stt-language`), `sendTextMessage` (`lk.chat`) and `speak` (`did.speak`) all call it now, and it is the single implementation forwarding to the internal `sendMessage(payload, topic)`. WebRTC's `interrupt` routes through its own implementation the same way. Wire behavior is unchanged - identical topics and identical payload strings; verified by the existing pins (`livekit-manager.test.ts` STT/data-message `sendText` assertions, `webrtc-core`/`edge-cases` `dataChannel.send` assertions). `yarn ci:test` 534/534, coverage 84.05/73.27/89.18/88.9 vs thresholds 80/70/86/87, prettier clean.

src/services/agent-manager/index.ts
```diff
@@ -351,6 +352,17 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
 
             return items.streamingManager.setSttLanguage(language);
         },
+        sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
```
dor-eitan: shouldnt we call it sendDataChannelMessage?

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -725,7 +726,7 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
         }
     }
 
-    async function sendMessage(message: string, topic: DataChannelTopic) {
+    async function sendMessage(message: string, topic: string) {
```
dor-eitan: shouldnt we keep the topic as DataChannelTopic?

src/types/entities/agents/manager.ts
```diff
@@ -317,6 +318,14 @@ export interface AgentManager {
      */
     setSttLanguage: (language: string) => Promise<void>;
 
+    /**
+     * Send a JSON payload to the agent over a data-channel topic
+     * Only available for Expressive (V4) agents
+     * @param topic - Data-channel topic to send on (see `DataChannelTopic`)
+     * @param payload - Plain object, serialized as JSON
+     */
+    sendDataMessage: (topic: DataChannelTopic, payload: Record<string, unknown>) => Promise<void>;
```
dor-eitan: should we call it "sendDataMessage" or "sendDataChannelMessage"?

src/services/streaming-manager/common.ts
```diff
@@ -9,7 +9,9 @@ export const createStreamingLogger = (debug: boolean, prefix: string) => (messag
  */
 export type StreamingManager<T extends CreateStreamOptions | CreateSessionV2Options> = {
     /**
-     * Method to send request to server to get clip or talk depending on payload
+     * Method to send request to server to get clip or talk depending on payload.
+     * Each implementation owns the transport details (V1 posts a stream request
+     * to the API; V2 sends the serialized payload on the `did.speak` topic).
```
dor-eitan: remove the implementation details

src/services/streaming-manager/common.ts
```diff
@@ -20,17 +22,18 @@ export type StreamingManager<T extends CreateStreamOptions | CreateSessionV2Opti
...
+     * through here. LiveKit sends it as a text stream on the given topic;
+     * WebRTC sends the payload as-is on the RTC data channel and ignores the
+     * topic, having no topic concept. Callers that need a specific message
+     * (chat, STT language, presentation, ...) call this with the matching
+     * topic rather than getting a method of their own; only `speak` and
+     * `interrupt` keep members, because their mechanics genuinely differ per
+     * transport, and whatever they put on the data channel goes through here.
+     * @param topic Data-channel topic to send on (see `DataChannelTopic`)
+     * @param payload The message payload to send, already serialized
      */
```
dor-eitan: trim

src/services/agent-manager/index.ts
```diff
@@ -343,13 +344,26 @@ export async function createAgentManager(agent: string, options: AgentManagerOpt
...
+            return Promise.resolve();
+        },
+        sendDataMessage(topic: DataChannelTopic, payload: Record<string, unknown>): Promise<void> {
+            if (!isStreamsV2 || !items.streamingManager) {
+                return Promise.reject(new Error('sendDataMessage is not available for this streaming manager'));
+            }
+
+            analytics.track('agent-data-message', { topic });
+
+            items.streamingManager.sendDataChannelMessage(topic, JSON.stringify(payload));
```
dor-eitan: isnt the sendDataChannelMessage is async method? i see it across all the pr

src/services/agent-manager/index.ts
dor-eitan: Renamed in `ecf6a0d` — `AgentManager.sendDataMessage` is now `sendDataChannelMessage`, the same name as the transport primitive it delegates to. One concept, one name at both layers: the app-facing method takes the object payload and stringifies it into the primitive. Renamed on the interface (`types/entities/agents/manager.ts`), the implementation, the tests and the README; the Mixpanel event name stays `agent-data-message` since analytics consumers already read it. agents-ui is updated in parallel (de-id/agents-ui#1105).

src/types/entities/agents/manager.ts
dor-eitan: `sendDataChannelMessage` — done in `ecf6a0d`. Same name at the app layer and the transport layer, since it is the same concept; the AgentManager method just stringifies the object payload into the primitive. The error message moved with it (`sendDataChannelMessage is not available for this streaming manager`), and the README/tests follow.

src/services/streaming-manager/livekit-manager.ts
dor-eitan: Yes — fixed in `3cc59cd`. The reason it was `string` is that `DataChannelTopic` was declared *inside* this file, so the shared type in `common.ts` and `AgentManager` could only reach it by importing from a service (the types→service edge you flagged earlier), which would have become a real cycle. The enum moved to `src/types/stream/stream.ts` next to `StreamEvents`, and the package root still exports it (now via `export * from './types'`), so the public surface is unchanged.

With it reachable, the topic is typed `DataChannelTopic` on the shared type, both implementations, and `AgentManager.sendDataChannelMessage`. One knock-on: V1's `interrupt` used to pass `StreamEvents.StreamInterrupt` as the topic; it now passes `DataChannelTopic.Interrupt` with a comment saying the topic is ignored there (WebRTC has no topic concept — the payload's `type` identifies the interrupt), so no invented union.

src/services/streaming-manager/common.ts
dor-eitan: Trimmed in `6167493`. `speak`'s doc lost the per-transport details (V1 posts a stream request / V2 sends on `did.speak`) — that belongs in the implementations, where it already is — and I trimmed `interrupt`'s matching lines in the same style, since it had the same kind of block.

src/services/streaming-manager/common.ts
dor-eitan: Trimmed in `6167493` — 11 lines down to 3: what it is (the single send primitive), how the topic is treated per transport (V2 routes on it, V1 ignores it), and the promise semantics. The prose about which members stay on the type and why is gone; the code says it.

src/services/agent-manager/index.ts
dor-eitan: Good catch — it is async, and this was a real regression I introduced in `5f065f2`. Fixed in `cc56144`.

LiveKit's implementation has always been async (it awaits `room.localParticipant.sendText`), but the shared type declared `void`, so two `await`s got dropped when the call sites moved: the V2 chat path's `await ...sendTextMessage(userMessage)` became a bare call, and `setSttLanguage` returned `Promise.resolve()` instead of the send's promise. Both callers resumed before the bytes left, and `setSttLanguage`'s documented "resolves when sent" contract was silently gone.

Now: the primitive is typed `Promise<void>` on the shared type, WebRTC's implementation is `async` to match (body stays synchronous), the V2 chat path awaits again, and `setSttLanguage` / `sendDataChannelMessage` return the send's promise. Two tests pin it — both methods stay pending until the send settles (they fail against the pre-fix code).

On severity, to be honest about it: the promise never rejected on main either — send failures go to `callbacks.onError`, and the doc now says so — so nothing was swallowing errors. What was lost was the contract and the timing, and that is what is restored.

omrizitrin-did (approved):
nice!

PR #463, opened 2026-08-27
fix: forward persistentChat to streams v2 session creation

ReutAtias3:
### Pull Request Type

🐛 BugFix

### Description

-   `persistentChat` only reached the v1 chat endpoint (`POST /agents/{id}/chat`). Expressive (LiveKit) agents create the chat as part of session creation, and `getAgentStreamV2Options` never populated `chat_persist` — so the backend always applied its `true` default, and customers passing `persistentChat: false` silently got persisted chats.
-   Maps `persistentChat` → `chat_persist` on the v2 session payload (`POST /v2/agents/{id}/sessions`), omitting the field when unset so both the v1 and v2 backend defaults stay exactly as they are today. Only an explicit value changes behavior.
-   Tests for `true` / `false` / omitted-when-unset on the v2 options builder, plus LiveKit-manager coverage that an explicit `false` survives the existing `?? true` default.

Behavior matrix:

| `persistentChat` | v1 (`talk`/`clip`) → `persist` | v2 (`expressive`) → `chat_persist` |
| ---------------- | ------------------------------ | ---------------------------------- |
| unset            | `false` (unchanged)            | `true` (unchanged)                 |
| `true`           | `true` (unchanged)             | `true` (**was** always `true`)     |
| `false`          | `false` (unchanged)            | `false` (**was** `true`)           |

The v1/v2 unset-default asymmetry is left in place on purpose — aligning them would change behavior for everyone who omits the flag — and is now documented on `AgentManagerOptions.persistentChat`.

### DIFF
--- a/src/services/agent-manager/connect-to-manager.ts
+++ b/src/services/agent-manager/connect-to-manager.ts
@@ -35,11 +35,12 @@ import {
 import { createChat } from '../chat';
 
 const ChatPrefix = 'cht';
-function getAgentStreamV2Options(): CreateSessionV2Options {
+function getAgentStreamV2Options(options?: ConnectToManagerOptions): CreateSessionV2Options {
     return {
         transport: {
             provider: TransportProvider.Livekit,
         },
+        ...(options?.persistentChat !== undefined && { chat_persist: options.persistentChat }),
     };
 }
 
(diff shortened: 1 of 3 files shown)


src/services/agent-manager/connect-to-manager.ts
```diff
@@ -35,11 +35,12 @@ import {
 import { createChat } from '../chat';
 
 const ChatPrefix = 'cht';
-function getAgentStreamV2Options(): CreateSessionV2Options {
+function getAgentStreamV2Options(options?: ConnectToManagerOptions): CreateSessionV2Options {
```
dor-eitan, line 38: do we want to allow all the options or Pick that specific one from there?

dor-eitan (approved):

src/services/agent-manager/connect-to-manager.ts
ReutAtias3: on v2 backend  (InitSessionV2Request type) we allow only chat_persist, and transport @dor-eitan

dor-eitan: but the interface will allow the consumer to pass anything from that object

PR #468, opened 2026-09-07
feat: report why the server ended the stream on the connection callback

osimhi213:
### Pull Request Type

🔮 Feature

### Description

- `onConnectionStateChange` now receives a second `reason` argument. On `disconnected` it is a `StreamEndReason` when the server ended the stream on purpose: the agent's end-call tool, inactivity, or a time or message limit. Otherwise it is an opaque transport diagnostic such as `livekit:disconnected`. Until now every server-side end looked identical to a dropped connection.
- `StreamEndReason` values: `ok`, `unknown_error`, `network_issue`, `message_limit`, `time_limit`, `inactivity`, `ended_by_agent`.
- No new callback. The SDK has always passed this reason internally and dropped it at the public boundary in `connect-to-manager`, so every disconnect gains a cause, not just this one. `agent-connection-state-change` already tracks the reason, so there is no new analytics event either.
- Expressive agents get the reason from the `stream/done` and `stream/error` data-channel events published by streaming-orchestrator-worker #734, which the manager attaches to the disconnect it already emits. Legacy agents get it from the session-level `stream/done` they already receive over the websocket, mapped from `close_reason` in `socket-manager/stream-end.ts`.
- Nothing about the connection behaviour changes. The agent leaving the room still ends it, and the manager neither disconnects itself nor blocks reconnects.
- The per-video `stream/done` on the legacy WebRTC data channel is untouched and still drives video state in `webrtc-manager.ts`.

### Reference Links

-   [Asana](https://app.asana.com/1/856614567666442/project/1217184354211990/task/1216916819996461?focus=true)


osimhi213:
Superseded by https://github.com/de-id/agents-sdk/pull/470, same tree squashed into one commit.

PR #470, opened 2026-09-07
feat: report why the server ended the stream

osimhi213:
### Pull Request Type

🔮 Feature

### Description

- `onConnectionStateChange` now receives a second `reason` argument. On `disconnected` it is a `StreamEndReason` when the server ended the stream on purpose, so an app can tell the agent hanging up, an inactivity timeout or a time or message limit from a dropped connection. Otherwise it is an opaque transport diagnostic such as `livekit:disconnected`.
- No new callback. The SDK has always passed this reason internally and dropped it at the public boundary, so every disconnect gains a cause, not just this one. The existing `agent-connection-state-change` event already tracks it.
- Expressive agents read it from the `stream/done` and `stream/error` data-channel events added in streaming-orchestrator-worker #734. Legacy agents read it from the session-level `stream/done` they already receive over the websocket, mapped from `close_reason`.
- Nothing about the connection behaviour changes, and the per-video `stream/done` on the legacy WebRTC data channel is untouched.

### Reference Links

-   [Asana](https://app.asana.com/1/856614567666442/project/1217184354211990/task/1216916819996461?focus=true)

### DIFF
--- a/src/services/streaming-manager/livekit-manager.ts
+++ b/src/services/streaming-manager/livekit-manager.ts
@@ -796,6 +796,7 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
 
             log('Reconnecting to LiveKit room, state:', room.state);
             hasEmittedConnected = false;
+            streamEndReason = null;
             callbacks.onConnectionStateChange?.(ConnectionState.Connecting, 'user:reconnect');
 
             try {
--- a/src/services/streaming-manager/livekit-manager.ts
+++ b/src/services/streaming-manager/livekit-manager.ts
@@ -9,6 +9,7 @@ import {
     Message,
     PayloadType,
     RunningToolCall,
+    StreamEndReason,
     StreamEvents,
     StreamingManagerOptions,
     StreamingState,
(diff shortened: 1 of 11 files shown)


src/services/streaming-manager/livekit-manager.ts
```diff
@@ -508,6 +513,10 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
         callbacks.onAgentActivityStateChange?.(AgentActivityState.Idle);
     }
 
+    function handleStreamEnded(_subject: string, data: { reason: StreamEndReason }): void {
```
dor-eitan, line 520: when do you clear that var?

dor-eitan (approved):

src/services/streaming-manager/livekit-manager.ts
osimhi213: > when do you clear that var?

Nowhere, which was a real bug: it lived as long as the manager, so a manager that reconnected after a deliberate end reported that same reason on the next disconnect regardless of the actual cause. Reachable today, because the UI keeps its reconnect screen for `inactivity` and its button calls `reconnect()`.

Fixed in 4234e1a by clearing it in `reconnect()`, beside the existing `hasEmittedConnected` reset, since it is per-connection state. Clearing it on `disconnect()` would break the feature, as the `Disconnected` event that reads it arrives afterwards. Added a test that fails without the reset.

_authored by claude using pr:review@de-id skill_

PR #472, opened 2026-09-08
fix: emit the first chat partial and never lose the first video event

dor-eitan:
### Pull Request Type

🐛 BugFix

### Description
-   **First partial of a new assistant message was never emitted.** `processChatEvent` created the message with the partial's content and then skipped `onNewMessage` because the content-changed check compared equal. Multi-token replies hid this (token two fires it); a single-partial message, such as a worker-side `say` greeting, never reached the UI until its `chat/answer` seconds later. The message is now created empty so the first partial emits. Consumers see one extra `onNewMessage('partial')` at the start of every assistant message.
-   **First `stream-video/started` of a session was lost.** `handleVideoEvents` read `getReport()` before anything else; `createVideoStatsReport` indexed `stats[0]` unguarded and threw before the first sample, and the data-channel `catch` only logged under `debug`. The event produced no `onMessage` and no `Talking` transition, so the UI kept hiding the first assistant message. `getReport()` now returns `null` with no samples and every caller is null-safe.
-   `handleDataReceived` parses and dispatches in separate `try` blocks; a handler exception is now logged through a new always-on `warn` channel on the streaming logger, so a swallowed handler failure is visible in production consoles with the `[LiveKitStreamingManager]` prefix. Parse failures keep the debug-gated log.
-   Found while validating de-id/streaming-orchestrator-worker#729, where the worker says the greeting itself; both bugs predate it.

### Reference Links
-   [Worker PR](https://github.com/de-id/streaming-orchestrator-worker/pull/729)

https://claude.ai/code/session_01CXAuKEuqTp6XtE7u8P9r6A

### DIFF
--- a/src/services/socket-manager/message-queue.ts
+++ b/src/services/socket-manager/message-queue.ts
@@ -94,7 +94,10 @@ function processChatEvent(
         currentMessage = {
             id: data.id || `assistant-${Date.now()}`,
             role: data.role || 'assistant',
-            content: data.content || '',
+            // Starts empty so the content-change check below fires for this message's first chat
+            // event. Seeding it with `data.content` made that check a no-op, so a message carried
+            // by a single partial (e.g. a worker `say` greeting) never reached `onNewMessage`.
+            content: '',
             parts: [],
             created_at: data.created_at || new Date().toISOString(),
         };
--- a/src/services/streaming-manager/common.ts
+++ b/src/services/streaming-manager/common.ts
@@ -1,24 +1,8 @@
 import { CreateSessionV2Options, CreateStreamOptions, Interrupt, PayloadType, StreamType } from '@sdk/types';
 import { DataChannelTopic } from '@sdk/types/stream/data-channel';
 
-/** A log channel. Accept this where only logging is needed - it fits `log` and `log.warn` both. */
-export type StreamingLog = (message: string, extra?: any) => void;
-
-export type StreamingLogger = StreamingLog & { warn: StreamingLog };
-
-export const createStreamingLogger = (debug: boolean, prefix: string): StreamingLogger => {
-    const log: StreamingLog = (message, extra) => {
-        debug && console.log(`[${prefix}] ${message}`, extra ?? '');
-    };
-
-    // The only always-on channel. Reserved for failures a consumer cannot otherwise observe -
-    // gating those behind `debug` is what let a swallowed handler exception go unnoticed.
-    const warn: StreamingLog = (message, extra) => {
-        console.warn(`[${prefix}] ${message}`, extra ?? '');
-    };
-
-    return Object.assign(log, { warn });
-};
+export const createStreamingLogger = (debug: boolean, prefix: string) => (message: string, extra?: any) =>
+    debug && console.log(`[${prefix}] ${message}`, extra ?? '');
 
 /**
  * Shared type for all streaming managers (LiveKit, WebRTC, etc.)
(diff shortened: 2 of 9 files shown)


src/services/socket-manager/message-queue.ts
```diff
@@ -94,7 +94,10 @@ function processChatEvent(
         currentMessage = {
             id: data.id || `assistant-${Date.now()}`,
             role: data.role || 'assistant',
-            content: data.content || '',
+            // Starts empty so the content-change check below fires for this message's first chat
+            // event. Seeding it with `data.content` made that check a no-op, so a message carried
+            // by a single partial (e.g. a worker `say` greeting) never reached `onNewMessage`.
```
dor-eitan, line 99: does it mean we wont get anything in the first content?

src/services/streaming-manager/livekit-manager.ts
```diff
@@ -547,18 +547,29 @@ export async function createLiveKitStreamingManager<T extends CreateSessionV2Opt
...
+        const handler = dataChannelHandlers[subject];
+        if (!handler) return;
+
+        try {
+            handler(subject, data);
         } catch (e) {
-            log('Failed to parse data channel message:', e);
+            // Always on: a handler exception drops the whole event (no onMessage, no state
+            // transition), and a debug-gated log made that symptom invisible in production.
+            log.warn('Data channel handler failed', { subject, error: e });
```
dor-eitan: just use console, remove that from the logger

src/services/streaming-manager/common.ts
```diff
@@ -1,8 +1,24 @@
...
+        debug && console.log(`[${prefix}] ${message}`, extra ?? '');
+    };
+
+    // The only always-on channel. Reserved for failures a consumer cannot otherwise observe -
+    // gating those behind `debug` is what let a swallowed handler exception go unnoticed.
+    const warn: StreamingLog = (message, extra) => {
+        console.warn(`[${prefix}] ${message}`, extra ?? '');
+    };
+
+    return Object.assign(log, { warn });
```
dor-eitan: unneeded complexity

src/services/socket-manager/message-queue.ts
dor-eitan: No. The message object is seeded empty only so the content-change check below is true for its first event. In the same call the partial is written into `chatEventQueue`, `getMessageContent` is computed, `currentMessage.content` is set to it, and `onNewMessage` fires — so the first callback already carries the full text of the first partial. Consumers never observe an empty message; before this change they observed nothing at all until the second partial (or the answer, for a single-partial `say`). Test (a) in the spec asserts the emitted content equals the partial text.

src/services/streaming-manager/livekit-manager.ts
dor-eitan: Done in 878ddd2: the dispatch catch calls `console.warn('[LiveKitStreamingManager] Data channel handler failed', { subject, error })` directly. The parse/dispatch split stays so a non-JSON payload keeps the debug-gated log and never warns.

src/services/streaming-manager/common.ts
dor-eitan: Done in 878ddd2: `createStreamingLogger` is back to its main form, the `StreamingLog`/`StreamingLogger` types and the `warn` channel are gone, and the `webrtc-manager.ts` annotations are restored.

osimhi213 (approved):
remove the comments

PR #478, opened 2026-09-15
docs: document every public symbol and gate on missing docs (API reference, part 2)

dor-eitan:
### Pull Request Type

🔮 Feature

### Description

Second of three PRs for the API reference at **https://sdk.d-id.com/**. Builds on #477 (3.0.0 public surface). This PR documents every public symbol in the source and turns the missing-documentation check into a CI gate, so the reference can no longer drift from the code.

- **Every public symbol documented** — classes, interfaces, enums, functions, type aliases, every member and parameter (TypeDoc coverage badge: 100%). Each statement was checked against its call site; server-side behaviour the SDK cannot verify is written as such. Avatar applicability is stated per member (Talks (V2) / Clips (V3) / Expressive (V4)).
- **Sidebar by concept** via `@category`: Guides, Agent Manager, Authentication, Callbacks & Events, Streaming Options, Speak & Scripts, Chat, Voice, Errors. The migration guide is filed under Guides through a small wrapper document so `MIGRATION.md` stays plain Markdown on GitHub.
- **Gate on.** `yarn docs:check` now fails on any undocumented public symbol, in addition to unexported references and broken `{@link}`s. CI runs it on every PR. The gate catches *missing* documentation; *wrong* documentation is caught by review, which is why each statement here was verified against its call site.
- **`StreamOptions` is filed under Streaming Options** (the spec listed it under Agent Manager); it sits next to the stream enums it configures.
- **README trimmed to a guide.** Method-by-method prose moved into the source; the README keeps intro, prerequisites, install, one end-to-end example and links into the reference. Fixed a wrong field path (`agent.idle_video`, not `agent.presenter.idle_video`) and three broken in-page anchors. Coverage badge restored.
- **One type narrowing.** `AgentManager.getSTTToken` is now `Promise<STTTokenResponse>`; the implementation never resolved `undefined` (a failed request throws `HttpError`). Recorded in the migration guide.
- **Project rules in `CLAUDE.md`.** The public-API rule (JSDoc + `@category` on every export, majors for removals, no `export *`) moved out of the `src/index.ts` header into a repo-level `CLAUDE.md`.

### Audit round

[...]

PR #479, opened 2026-09-15
fix: defects surfaced by the API-reference pass (429 retry, interrupt, chat modes, callbacks)

dor-eitan:
### Pull Request Type

🐛 BugFix

### Description

Code fixes for defects found while writing the API reference (#478). Stacked on the docs branch so the affected doc comments and `MIGRATION.md` are updated in the same commits. All behaviour changes ship in 3.0.0.

- **429 responses are now retried.** The client's retry wrapper (three attempts, one second apart) sat around `fetch()`, which resolves on a 429, so it never fired. The retried operation now throws on a 429 and the final response goes through the normal `HttpError` path. Replaces the June 2026 test that pinned the non-retrying behaviour; if that was intentional, say so and this commit can be dropped.
- **`interrupt()` never throws.** On fluent Talks/Clips streams the WebRTC manager threw when no video was playing, after the SDK had already marked the last message `interrupted`. The transports now report whether an interrupt was actually sent, return silently otherwise, and the message is marked only when one was. `isInterruptible` keeps its meaning (interrupting currently allowed, e.g. not while a blocking client tool runs); `getIsInterruptAvailable()` remains the session capability.
- **`persistentChat` defaults to `false` on every agent type.** Expressive sessions were created with persistence on unless `false` was passed; Talks/Clips defaulted off. One default now.
- **`Message.videoId` removed.** The SDK never populated it.
- **`SDK_VERSION` no longer exported.** Internal analytics value; nothing outside the SDK read it.
- **`AgentManagerCallbacks` declared uniformly** as function-typed properties, so the reference renders one list instead of "Properties" and "Methods". Type-neutral for callers passing object literals.
- **`ChatCreationFailed` message** no longer has a double space when `persistent` is false.
- **`yarn docs:serve`** builds the reference and serves it at http://localhost:4173 for local review.
- **`changeMode()` returns `Promise<void>`.** It was always asynchronous (it may disconnect the stream) but was typed `void`, which made its errors uncatchable. `connect()` now awaits the mode it applies and ignores an unsupported server-provided mode with a warning instead of failing silently.
- **`isInterruptible` on Expressive is live**, not a creation-time snapshot, so `interrupt()` is correctly held back while a blocking client tool runs.
- **`ChatMode.Off` and `DirectPlayback` rejected on Expressive agents.** They were never supported there; `createAgentManager` and `changeMode()` now throw `ValidationError` instead of running a half-configured session.

- **`kind` is a literal per error class** (`'StreamError'`, `'WSError'`, …), so `BaseError` narrows on it and each reference page shows the real value. `HttpError.kind` stays `string` (it carries the server's classification).
- **`rate()`, `deleteRate()`, `submitFeedback()` are `async`**, so their `ValidationError` rejects instead of throwing synchronously.
- **Wrong-avatar-type guards reject with `ValidationError`** (`publishMicrophoneStream`, `replaceMicrophoneTrack`, `publishCameraStream`, `setSttLanguage`, `sendDataChannelMessage`) instead of a plain `Error`, so `isDIDError` recognises them.
- **`HttpError.endpoint`** replaces `HttpError.url`, matching `NetworkError` and the `toJson()` key; the value is a path relative to the API base.
- **`MessageSentiment`** names the previously anonymous `Message.sentiment` shape.
- **`initialMessages` get their `parts` filled** from `content` when missing or empty, and the SDK no longer pushes into the array you passed.

### Consumers

- `interrupt()` callers that wrapped it in `try/catch` for the idle case can drop the guard.
- `changeMode()` callers should `await` it (and may catch `ValidationError` on Expressive).
- Expressive apps that relied on the implicit persistent chat must now pass `persistentChat: true`.
- Expressive apps passing `mode: ChatMode.Off` or `DirectPlayback` will now fail at creation; use `TextOnly`/`Functional`.
- Code matching on the message `is not available for this streaming manager` should catch `ValidationError` instead.

### Verification

`yarn type-check`, `yarn lint`, `yarn test:ci`, `yarn docs:check`, `yarn docs:build`; see the PR checks.

### Reference Links
-   [Asana](https://app.asana.com/1/856614567666442/project/1217823176333868/task/1217777263616315?focus=true)

https://claude.ai/code/session_01JAmXKgcXozg5TwvJDiUtDf


MIGRATION.md
```diff
@@ -45,11 +45,17 @@ Shapes are unchanged; only the names differ.
...
 - `StreamEvents.StreamCreated` — never emitted; use the `onStreamCreated` callback.
+- `Message.videoId` — never set by the SDK.
+- `SDK_VERSION` — internal analytics value; no longer exported.
 
 ## Behaviour clarifications
 
 - `speak()` on Expressive (V4) agents now resolves with `{ status: 'success', duration: 0, video_id: '' }` instead of `undefined`, matching its declared type.
 - `agentManager.getSTTToken()` is now typed `Promise<STTTokenResponse>`; it never resolved `undefined` (a failed request throws `HttpError`).
+- `interrupt()` never throws; where it used to throw on Talks (V2)/Clips (V3) streams it now returns silently, and the last message is marked `interrupted` only when an interrupt was actually sent.
+- `persistentChat` now defaults to `false` on Expressive (V4) agents too; v2 sessions were created with chat persistence on unless you passed `false`.
```
ReutAtias3: remove the now.. just default false for all

MIGRATION.md
```diff
@@ -45,11 +45,17 @@ Shapes are unchanged; only the names differ.
 - `AgentManagerOptions.microphoneStream` — it was never read by the SDK, so passing it had no effect. Call `agentManager.publishMicrophoneStream(stream)` after `connect()` instead (Expressive (V4) agents).
 - The Expressive-only media methods (`publishMicrophoneStream`, `unpublishMicrophoneStream`, `replaceMicrophoneTrack`, `publishCameraStream`, `unpublishCameraStream`) are now required members of `AgentManager` instead of optional. They always existed at runtime; on Talks (V2) and Clips (V3) agents the `publish`/`replace` methods reject and the `unpublish` methods resolve without effect. Remove any `?.` guards.
 - `StreamEvents.StreamCreated` — never emitted; use the `onStreamCreated` callback.
+- `Message.videoId` — never set by the SDK.
```
ReutAtias3: if its never set, we can just not mention, no ?

ReutAtias3 (approved):

MIGRATION.md
dor-eitan: Now: "`persistentChat` defaults to `false` for all agent types."

MIGRATION.md
dor-eitan: Removed the bullet.

dor-eitan: Follow-up: a whole-reference audit flagged that dropping the line hides a compile error for anyone reading `message.videoId` (the field is gone in 3.0), so the guide keeps one short line pointing at `ChatResponse.videoId` instead.

MIGRATION.md
dor-eitan: Follow-up: kept it to one line, but the audit pointed out the sentence read as a fact rather than a change, so it now says the default flipped for Expressive (V4) and that `persistentChat: true` keeps the v2 behaviour.

PR #484, opened 2026-09-16
fix!: release hardening for 3.0.0 — importable everywhere, exports map, narrowing errors, lazy LiveKit

dor-eitan:
### Pull Request Type

🐛 BugFix · 📦 Packaging — part of the **3.0.0** release (stacked on #483)

### Description

Release blockers found by a packaging and API review of the 3.0.0 candidate. Every consumer-visible change has one line in `MIGRATION.md`.

- **Importable everywhere.** Nothing browser-only runs at import time any more; the package loads under Node and SSR frameworks (was `ReferenceError: window is not defined`).
- **The build cannot silently fail.** `yarn build` exits non-zero on any failure (it exited 0).
- **`exports` map, npm metadata, LICENSE.** `types` / `import` / `require` conditions, `sideEffects: false`, `engines`, `homepage`, `bugs`, description and keywords; test declarations no longer ship; tarball 116 → 9 files.
- **Declarations match the reference.** `stripInternal` on, one bundled `dist/index.d.ts` (works under `moduleResolution: node16`, which previously resolved the whole SDK to `any`); `baseURL`, `wsURL`, `enrichAnalytics()` and `Agent.ld_context` are public, documented as advanced; `vm_account_id` is `@hidden` (typed, not in the reference). Rule recorded in `CLAUDE.md`.
- **The error model narrows.** `HttpError.kind` is the literal `'HttpError'`; the Agents API's classification moves to `HttpError.code`; `DIDError` union exported; `isDIDError()` narrows to it, so `switch (e.kind)` is exhaustive.
- **Mode guards read the current mode**, not the creation-time one; a mode switch reconnects what the new mode needs.
- **Options are read, never mutated**; **`connect()` is idempotent** (in-flight promise reused, second connect on a live session rejects with `ValidationError`, `disconnect()` during a connect is honoured).
- **Tool-call payloads agree** with `RunningToolCall`; the error payload carries the failure reason.
- **LiveKit loads only for Expressive (V4) agents.** ESM entry 186 KB → 17 KB gzip; LiveKit is a lazy chunk. ESM build minified.
- README states above the fold that 3.0 is breaking.

### Consumers

agents-ui companion on de-id/agents-ui #1136 (`HttpError.code`, tool payload fields).

### Verification

`yarn type-check`, `yarn lint`, `yarn test:ci` (635), `yarn docs:check` (zero warnings), `yarn build`; consumer type-check under `bundler` and `node16`; `npm pack --dry-run`.

### Reference Links
-   [Asana](https://app.asana.com/1/856614567666442/project/1217823176333868/task/1217777263616315?focus=true)

https://claude.ai/code/session_01JAmXKgcXozg5TwvJDiUtDf


dor-eitan:
@ReutAtias3 one thing in here reverses a hide you asked for on #478, so flagging it explicitly.

`baseURL`, `wsURL`, `enrichAnalytics()` and `Agent.ld_context` are public again, documented as advanced ("used by D-ID's own applications; not part of the SDK contract"). Reason: this PR turns on `stripInternal`, so anything tagged `@internal` disappears from the shipped `.d.ts` as well as from the site. agents-ui reads all four, and a first-party app depending on a member makes it part of the surface whether or not the reference prints it. Hiding it in one place while shipping it in another was the inconsistency we were trying to remove.

`advanced_settings.vm_account_id` stays out of the reference (Dor's call) via `@hidden`, which keeps it typed but unrendered. The `@internal` vs `@hidden` rule is now written in `CLAUDE.md`.

